PR #144, opened 2026-07-04
fix(loop-context): reject invalid numeric limit flags

### LINKED ISSUE #139: loop-context accepts invalid numeric limits and can fail open
I was trying to use `loop-context --check` as a guard in a small loop script and noticed that bad numeric limit values are accepted silently.

Repro:

```bash
printf '%s' '{"goal":"x","attempts":[{"iteration":1,"action":"a","outcome":"failure","error":"boom"},{"iteration":2,"action":"a","outcome":"failure","error":"boom"},{"iteration":3,"action":"a","outcome":"failure","error":"boom"}]}' | node tools/loop-context/dist/cli.js --check --json
```

That exits 2 with `trigger: "stagnation"`, which is what I expected for the default threshold of 3.

With an invalid value:

```bash
printf '%s' '{"goal":"x","attempts":[{"iteration":1,"action":"a","outcome":"failure","error":"boom"},{"iteration":2,"action":"a","outcome":"failure","error":"boom"},{"iteration":3,"action":"a","outcome":"failure","error":"boom"}]}' | node tools/loop-context/dist/cli.js --check --stagnation nope --json
```

it exits 0 and returns `trigger: "ok"`.

I expected `--stagnation nope` to fail with a usage/config error, since this flag controls when the loop stops. Instead, the invalid value gets through and the breaker continues. I see the same shape with other numeric limit flags such as `--no-progress` and `--max-iterations`.

The practical risk is a typo in a wrapper script disabling a safety limit instead of failing fast.

cobusgreyling:
Fixes #139

## Problem
Invalid numeric limit flags (`--stagnation nope`, `--no-progress 0`, etc.) were parsed with `Number()` and became `NaN`, which silently disabled the circuit breaker instead of failing closed.

## Fix
- Validate all numeric CLI flags as positive integers before running any operation
- Exit **1** with a clear usage/config error when a flag value is missing or invalid
- Add CLI integration tests covering the reported repro

## Repro (now fails as expected)
\`\`\`bash
printf '%s' '{"goal":"x","attempts":[...]}' | node tools/loop-context/dist/cli.js --check --stagnation nope --json
# loop-context failed: --stagnation must be a positive integer; got "nope".
# exit 1
\`\`\`

### DIFF
--- a/tools/loop-context/dist/cli.js
+++ b/tools/loop-context/dist/cli.js
@@ -1,6 +1,17 @@
 #!/usr/bin/env node
 import { readFile } from 'node:fs/promises';
 import { buildContextInjection, checkCircuitBreaker, pruneLedger, summarizeAttempts, DEFAULT_BREAKER, DEFAULT_PRUNE, } from './context-manager.js';
+/** Reject NaN/0/floats so a bad flag cannot silently disable the breaker. */
+function parsePositiveIntFlag(raw, flag) {
+    if (raw === undefined || raw === '') {
+        throw new Error(`${flag} requires a positive integer value.`);
+    }
+    const n = Number(raw);
+    if (!Number.isInteger(n) || n < 1) {
+        throw new Error(`${flag} must be a positive integer; got "${raw}".`);
+    }
+    return n;
+}
 function parseArgs(argv) {
     const breaker = { ...DEFAULT_BREAKER };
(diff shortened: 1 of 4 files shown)


Zhang-986:
Thanks for the quick fix. This covers the repro I hit when wiring `loop-context --check` into a small wrapper script.

PR #165, opened 2026-07-06
feat(loop-guard): size the token budget from loop-cost per pattern

KhaiTrang1995:
## Summary

Follow-up to the circuit breaker landed in #140. The breaker's `--token-budget`
was left to a hand-typed number, so in practice fix loops ran with no token cap
at all. `loop-cost` already computes a realistic per-run estimate per
pattern/level — this wires the two together so the breaker trips on real cost
blowup instead of a guess.

## What

- `loop-init` seeds `loop-ledger.json` with the loop's `pattern` and `level`
  (extra fields the breaker ignores) so `loop-guard` knows what to price.
- `loop-guard` skill derives `--token-budget` from `loop-cost --json`
  (`scenarios.realistic.tokensPerRun`).
- Cross-linked from `loop-cost`'s README.

## Why this shape

Pure shell wiring — no code coupling, no new dependency. `loop-context` stays
dependency-free, and the two tools remain independently usable, matching the
repo's "independent packages, wire at the edges" convention.

## Tested

- 12/12 `loop-init` tests (new assertion: seeded ledger carries a valid
  pattern + L1/L2/L3 level); typecheck clean.
- End-to-end: scaffold `ci-sweeper` → budget resolves to 19,250 → clean ledger
  returns CONTINUE (exit 0); a ledger over budget returns
  `ESCALATE [token-budget]` (exit 2).

### DIFF
--- a/templates/SKILL.md.loop-guard
+++ b/templates/SKILL.md.loop-guard
@@ -18,11 +18,15 @@ The breaker needs no LLM call, so it is cheap enough to run on every iteration.
 
 ## The ledger
 
-`loop-ledger.json` records the loop's goal and one entry per attempt:
+`loop-ledger.json` records the loop's goal, its pattern/level, and one entry per
+attempt:
 
 ```json
-{ "goal": "Get failing CI green", "attempts": [] }
+{ "goal": "Get failing CI green", "pattern": "ci-sweeper", "level": "L2", "attempts": [] }
 ```
+
+`pattern` and `level` are seeded by `loop-init`; the breaker ignores them but the
+budget step below reads them to size `--token-budget` from real cost data.
 
 After every iteration, append what you just tried:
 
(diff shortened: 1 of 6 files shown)


cobusgreyling (approved):
Clean follow-up to #140 — wires the two tools at the skill/shell layer without coupling packages, which matches the repo convention.

**Verified locally:**
- 12/12 `loop-init` tests pass (ledger seeds `pattern` + `level`)
- `loop-cost --json` resolves `scenarios.realistic.tokensPerRun` (ci-sweeper/L2 → 19,250)
- Skill doc and README cross-links are accurate

Nice work on the LEDGER_LEVEL defaults (L2 for fix-capable, L1 for report-only). Merging.

cobusgreyling:
Merged — thanks @KhaiTrang1995! This closes the loop on #140's token-budget gap. The shell-wiring approach keeps loop-context dependency-free while giving fix loops a real cost ceiling.

PR #256, opened 2026-07-11
feat(loop-cost): model multi-agent orchestration cost on the action path

KhaiTrang1995:
## Summary

`loop-cost` estimates spend assuming one implementer pass per run. But loop
engineering's maker/checker split -- and multi-agent orchestration more broadly
(an independent verifier, parallel candidate fan-out, debate rounds) -- make the
action path cost several times that. Today users guess that multiplier by hand,
which is exactly the estimation this tool exists to remove.

## What

A new `--orchestration` / `-o` flag applies a multiplier to the **action path
only** -- the no-op scan and the single triage pass are untouched:

| Mode | Multiplier | Models |
|------|-----------|--------|
| `single` (default) | 1x | One implementer pass, self-checked |
| `maker-checker` | 2x | Implementer + an independent verifier pass (the L2+ default) |
| `parallel:N` | N+1 | N candidate agents fan out, plus a merge/arbiter pass |
| `debate:R` | 1+R | One proposer plus R critique-and-revise rounds |

The scaled action cost flows into the realistic blend, so the realistic per-run
cap that feeds `loop-context`'s circuit breaker reflects the real run shape.
Multipliers above 2x emit a warning -- deep fan-out is cheap to enable and
expensive to run unattended.

## Why this shape

- Single-purpose flag on the existing estimator, no new dependency, no registry
  change -- orchestration is a property of how you run a pattern, not of the
  pattern itself.
- Default `single` is 1x, so every existing estimate and test is unchanged.

## Tested

`npm test` green (12/12; 5 new tests: multiplier mapping, bad-spec rejection,
action-only scaling, default no-op, deep-fan-out warning). README documents the
modes. Version 1.0.3 -> 1.1.0

### DIFF
--- a/tools/loop-cost/README.md
+++ b/tools/loop-cost/README.md
@@ -27,6 +27,7 @@ npm test
 | `--pattern` | Pattern id (see `--list`) |
 | `--cadence` | Override cadence (e.g. `15m`, `1d`) |
 | `--level` | `L1`, `L2`, or `L3` (default `L1`) |
+| `--orchestration` | Multi-agent action cost: `single` (default), `maker-checker`, `parallel:N`, `debate:R` |
 | `--conservative` | Use slower cadence from ranges |
 | `--json` | Machine-readable output |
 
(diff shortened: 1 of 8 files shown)


cobusgreyling:
Thanks @KhaiTrang1995 — CI is green. Leaving open for a focused maintainer review of the orchestration cost model before merge; #258 stacks on this branch so we'll aim to land #256 first, then rebase #258.

cobusgreyling:
Merged — thanks @KhaiTrang1995! Clean design: action-path-only multiplier, default unchanged, good test coverage. Published as loop-cost v1.1.0.

PR #258, opened 2026-07-11
feat(loop-context): resolve token budget from loop-cost's pattern estimate

KhaiTrang1995:
## Summary

`loop-context`'s circuit breaker only accepts `--token-budget` as a hand-typed
number, and `loop-budget.md`'s "On budget exceed" section admits the budget
number itself is usually a guess. `loop-cost` already computes a realistic
per-run token estimate for every pattern and readiness level from
`patterns/registry.yaml` -- nothing connected the two.

## What

Add `--budget-from-pattern <id>` to `loop-context`'s CLI, plus `--budget-level`,
`--budget-scenario`, `--budget-cadence`, and `--budget-conservative` to shape
the lookup. It shells out to `loop-cost`'s built CLI -- the same
monorepo-sibling-then-installed-dependency resolution `loop-init` already uses
for `loop-audit` -- so the two packages stay independent at the source level:

- An explicit `--token-budget` always overrides the derived value.
- An unknown pattern id surfaces `loop-cost`'s own error message, not a silent
  fallback.

Also updates the `loop-guard` skill template and `loop-cost`'s README, both of
which previously worked around the missing wiring with a manual JSON-parsing
shell pipe (`BUDGET=$(... | node -e '...')`) -- that hack is no longer needed
now that the flag does it directly.

## Why this shape

- Mirrors the existing cross-tool integration pattern in this monorepo
  (subprocess to a built CLI, not a shared library import) rather than
  introducing a new dependency style.
- `loop-cost` has no dependency on `loop-context`, so no circular dependency.

## Tested

`tools/loop-context`: 31/31 passing after a fully clean rebuild (wiped
`node_modules`/`dist` in both `loop-context` and `loop-cost`, reinstalled,
rebuilt). 13 new tests: resolver unit tests (scenario selection, unknown
pattern, invalid scenario rejected before spawning) and CLI integration tests
(budget resolved from pattern, explicit `--token-budget` overriding, unknown
pattern surfacing `loop-cost`'s error). All pre-existing tests pass unmodified.
Version 1.0.0 -> 1.1.0

### DIFF
--- a/templates/SKILL.md.loop-guard
+++ b/templates/SKILL.md.loop-guard
@@ -38,30 +38,27 @@ After every iteration, append what you just tried:
 `outcome` is `success | failure | noop`. Always include `error` on failures —
 that is how the breaker detects a repeated (stagnant) failure.
 
-## Size the token budget from the pattern (once per run)
+## Size the token budget from the pattern
 
-Don't hand-type a token cap — derive it from the pattern's realistic per-run
-cost so the breaker trips on genuine cost blowup, not a made-up number.
-[`loop-cost`](https://github.com/cobusgreyling/loop-engineering/tree/main/tools/loop-cost)
-already computes this from `patterns/registry.yaml`; read the loop's `pattern`
-and `level` straight from the ledger:
+Don't hand-type a token cap — `loop-context` can derive it directly from the
+pattern's realistic per-run cost
+([`loop-cost`](https://github.com/cobusgreyling/loop-engineering/tree/main/tools/loop-cost)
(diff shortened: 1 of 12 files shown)


cobusgreyling:
Thanks @KhaiTrang1995 — CI is green. This PR depends on the loop-cost orchestration changes in #256; we'll review/merge #256 first, then rebase this branch onto main. No action needed from you until #256 lands.

cobusgreyling:
Merged on top of #256 — thanks @KhaiTrang1995! `--budget-from-pattern` wiring looks solid; 31/31 tests green after rebase. Published as loop-context v1.1.0.

PR #370, opened 2026-07-23
feat: introduce loop-sandbox for isolated air-gap execution

THRISHAL12345:
## Summary
Introduces `loop-sandbox`, an invisible air-gap that allows unattended agents to run commands inside isolated git worktrees, capturing all modifications into a reviewable `.patch` file before automatically destroying the sandbox.

## Changes
- [x] Tool change (loop-sandbox)

- [x] All required sections present for patterns
- [x] Links work from README, patterns/README, starters/README, docs/index
- [x] `STATE.md*` examples use `.example` suffix
- [x] Safety-related content references `docs/safety.md`
- [x] Ran `node tools/loop-audit/dist/cli.js .` (or on the starter) and addressed findings

## Testing / Dogfood
- [x] `loop-audit` passes on affected starters or this repo
- [x] Manual review of generated state / skill output
- [x] Included unit tests in `tools/loop-sandbox/test/sandbox.test.mjs` verifying cross-platform isolation and patch generation

```text
npx @cobusgreyling/loop-sandbox run -- node script.js

📦 Creating loop-sandbox: sandbox-beeeabcb
🚀 Executing inside sandbox: node script.js

🔍 Scanning sandbox for changes...
✅ Sandbox changes captured to patch: .loop-sandbox/patches/sandbox-beeeabcb.patch
🧹 Cleaning up sandbox worktree...

==================================================
🎉 Run completed. Changes isolated successfully!
To review the changes, run:
  npx @cobusgreyling/loop-sandbox review

To apply the changes immediately, run:
  git apply .loop-sandbox/patches/sandbox-beeeabcb.patch
==================================================

### CODE REVIEW

File: tools/loop-sandbox/src/sandbox.ts [hunk 1]
```diff
@@ -0,0 +1,145 @@
...
+    console.error(`❌ Failed to extract patch from sandbox:`, err);
+  }
+
+  // 5. Cleanup the worktree
+  console.log(`🧹 Cleaning up sandbox worktree...`);
+  try {
+    // Force remove it because it has uncommitted changes (which we already saved to a patch)
+    await cleanupWorktrees({
+      root,
+      statuses: ['active'],
```
cobusgreyling: **[bug]** Cleanup calls `cleanupWorktrees({ root, statuses: ['active'], force: true })`, which force-removes **every** active entry in `.loop-worktrees/manifest.json`, not only the sandbox just created. Concurrent or pre-existing loop-worktree attempts (CI sweeper, other agents) will be destroyed mid-flight. This also contradicts `loop-worktree`'s documented invariant that `active` is never swept automatically.

**Suggestion:** Remove only this run's worktree (e.g. `git worktree remove --force ${entry.path}`, then drop that id from the manifest, or extend `cleanupWorktrees` with a `runIds`/`ids` filter). Prefer mark-then-remove for this `runId` only; never pass a blanket `statuses: ['active']` with `force: true`.

tools/loop-sandbox/src/sandbox.ts
```diff
@@ -0,0 +1,145 @@
...
+      hasChanges = true;
+      patchFilePath = path.join(patchesDir, `${runId}.patch`);
+      const diffOutput = await git(['diff', '--cached', '--binary'], worktreeAbsPath);
+      await writeFile(patchFilePath, diffOutput, 'utf8');
+      console.log(`✅ Sandbox changes captured to patch: ${patchFilePath}`);
+    } else {
+      console.log(`ℹ️ No changes were made in the sandbox.`);
+    }
+  } catch (err) {
+    console.error(`❌ Failed to extract patch from sandbox:`, err);
```
cobusgreyling: **[bug]** If patch extraction fails (`git add` / `diff` throw), the error is logged and execution falls through to force cleanup. Uncommitted sandbox work is then permanently destroyed with no patch written and no non-zero failure for that condition—silent data loss for the primary purpose of the tool.

**Suggestion:** Wrap create → run → extract → cleanup carefully: on extract failure, skip force-remove (or mark the worktree `stale`/`rejected` and leave it on disk), rethrow or return a clear error so the user can recover from `.loop-worktrees/<runId>`.

tools/loop-sandbox/src/sandbox.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  const worktreeAbsPath = path.resolve(root, entry.path);
+  
+  console.log(`🚀 Executing inside sandbox: ${command} ${args.join(' ')}`);
+
+  // 3. Execute the user's command
+  let exitCode: number | null = null;
+  try {
+    const child = spawn(command, args, {
+      cwd: worktreeAbsPath,
```
cobusgreyling: **[bug]** Isolation is only `cwd: worktreeAbsPath`. The process retains full filesystem, network, and shared-repo git access. From `.loop-worktrees/<id>`, paths like `../../package.json` write into the main working tree; absolute paths, `git -C`, push/force-push, and credential use are unrestricted. README/help call this an "air-gap" / "isolated" sandbox, which is inaccurate and dangerous if operators rely on it for unattended L2.

**Suggestion:** Document hard limits (cwd-only isolation; not a container/seccomp air-gap). Soften marketing language to "ephemeral worktree + patch capture." Point users at real isolation (containers, bubblewrap, OS sandboxes) for hostile agents, and link `docs/safety.md`.

tools/loop-sandbox/src/sandbox.ts
```diff
@@ -0,0 +1,145 @@
...
+    await git(['add', '-A'], worktreeAbsPath);
+    
+    // Check if there are any staged changes
+    const diffStat = await git(['diff', '--cached', '--stat'], worktreeAbsPath);
+    
+    if (diffStat) {
+      hasChanges = true;
+      patchFilePath = path.join(patchesDir, `${runId}.patch`);
+      const diffOutput = await git(['diff', '--cached', '--binary'], worktreeAbsPath);
+      await writeFile(patchFilePath, diffOutput, 'utf8');
```
cobusgreyling: **[bug]** Patch content from `git diff --cached --binary` is written with `writeFile(..., 'utf8')`. Binary-diff output can contain non-UTF-8 byte sequences; UTF-8 encoding can corrupt the patch so `git apply` fails or applies garbage.

**Suggestion:** Write the raw buffer: use `encoding: 'buffer'` in `execFile` and write the buffer directly without `.trim()`, or `await writeFile(patchFilePath, Buffer.from(diffOutput, 'binary'))` only if the string path is unavoidable.

tools/loop-sandbox/src/sandbox.ts
```diff
@@ -0,0 +1,145 @@
...
+  runId: string;
+  patchFile: string | null;
+  exitCode: number | null;
+  hasChanges: boolean;
+}
+
+/** Run a git command in cwd, returning stdout. Throws on error. */
+async function git(args: string[], cwd: string): Promise<string> {
+  const { stdout } = await runExec('git', args, { cwd, maxBuffer: 50 * 1024 * 1024 });
+  return stdout.trim();
```
cobusgreyling, line 25: **[bug]** `git()` always does `stdout.trim()`. For multi-file / binary patches, trimming leading/trailing whitespace can alter patch content (final newlines and edge bytes). Combined with string decoding, this weakens patch fidelity.

**Suggestion:** For patch capture specifically, use a raw `execFile` helper that returns untrimmed `Buffer`/`string`. Keep `trim()` only for human-facing status commands like `--stat`.

tools/loop-sandbox/src/sandbox.ts
```diff
@@ -0,0 +1,145 @@
...
+  
+  // 1. Setup paths
+  const sandboxDir = path.join(root, '.loop-sandbox');
+  const patchesDir = path.join(sandboxDir, 'patches');
+  await mkdir(patchesDir, { recursive: true });
+
+  console.log(`\n📦 Creating loop-sandbox: ${runId}`);
+  
+  // 2. Create the worktree
+  const entry = await createWorktree({
```
cobusgreyling, line 59: **[bug]** `createWorktree` is called without `base`, so `loop-worktree` defaults to branch `main`. Repos whose default branch is `master`, `trunk`, or a feature HEAD fail worktree creation. The sandbox also always starts from that base tip, not the caller's current HEAD/dirty tree—surprising and undocumented.

**Suggestion:** Pass `base` from CLI (e.g. `--base`), defaulting to `git rev-parse --abbrev-ref HEAD` or `git symbolic-ref refs/remotes/origin/HEAD`. Document that the sandbox is a clean checkout of `base`, not a copy of the dirty working tree.

tools/loop-sandbox/src/sandbox.ts
```diff
@@ -0,0 +1,145 @@
...
+    }
+  } catch (err) {
+    console.error(`❌ Failed to extract patch from sandbox:`, err);
+  }
+
+  // 5. Cleanup the worktree
+  console.log(`🧹 Cleaning up sandbox worktree...`);
+  try {
+    // Force remove it because it has uncommitted changes (which we already saved to a patch)
+    await cleanupWorktrees({
```
cobusgreyling: **[bug]** After `git worktree remove`, the branch `loop/<runId>` created by `createWorktree` is left behind. Repeated sandbox runs accumulate orphan branches indefinitely; there is no `git branch -D` / equivalent.

**Suggestion:** After successful worktree removal, delete the local branch (`git branch -D loop/${runId}` or `git update-ref -d refs/heads/loop/${runId}`), and document that remote tracking is not used.

tools/loop-sandbox/src/sandbox.ts
```diff
@@ -0,0 +1,145 @@
...
+  
+  console.log(`🚀 Executing inside sandbox: ${command} ${args.join(' ')}`);
+
+  // 3. Execute the user's command
+  let exitCode: number | null = null;
+  try {
+    const child = spawn(command, args, {
+      cwd: worktreeAbsPath,
+      stdio: 'inherit',
+      shell: true // support complex shell commands if needed
```
cobusgreyling: **[bug]** Child is spawned with `shell: true` while also passing an `args` array. Node docs warn that `shell: true` must not receive unsanitized input; argument joining is platform-specific and breaks paths with spaces. For a security-framed tool this widens injection surface without clear benefit when the CLI already accepts `run -- <cmd> args...`.

**Suggestion:** Default `shell: false` for argv-style execution; add an explicit `--shell` flag for users who need `bash -c`. Document the tradeoff.

tools/loop-sandbox/src/sandbox.ts
```diff
@@ -0,0 +1,145 @@
...
+    pattern: 'sandbox', // dummy pattern
+  });
+
+  const worktreeAbsPath = path.resolve(root, entry.path);
+  
+  console.log(`🚀 Executing inside sandbox: ${command} ${args.join(' ')}`);
+
+  // 3. Execute the user's command
+  let exitCode: number | null = null;
+  try {
```
cobusgreyling: **[bug]** No `SIGINT`/`SIGTERM` handlers and no `try/finally` around create→cleanup. Killing the CLI during a long agent run leaves the worktree, branch, and manifest entry as `active`. Cleanup errors are only logged; return value of `cleanupWorktrees` is ignored, so partial failures look successful.

**Suggestion:** Register signal handlers that force-remove this `runId` only; always `try/finally` cleanup for this entry; check `removed`/`skipped` and exit non-zero if the sandbox was not removed; print recovery instructions (path + branch).

tools/loop-sandbox/README.md
```diff
@@ -0,0 +1,65 @@
+# loop-sandbox
+
+The ultimate L2 safety mechanism for Loop Engineering. Divert agent execution into an ephemeral git worktree, capture their changes as a reviewable `.patch` file, and instantly destroy the sandbox.
+
+## Why loop-sandbox?
+
+Unattended AI agents (loops) running directly in your main working tree can cause damage, overwrite untracked files, or make it incredibly tedious to untangle failed execution attempts.
+
+`loop-sandbox` solves this by acting as an invisible air-gap:
```
cobusgreyling: **[suggestion]** README markets an "invisible air-gap" and claims capture of "*everything* the agent did (including new untracked files)". Implementation cannot air-gap network or FS, and `git add -A` skips gitignored paths, so ignored artifacts (build outputs, `.env` if ignored, tool caches) are not in the patch.

**Suggestion:** Replace "air-gap" with accurate wording (ephemeral worktree isolation for relative edits). Document exclusions: gitignored files, changes outside the worktree, git config/ref operations. Reference `docs/safety.md` and state this is complementary to denylist/gate tools, not a substitute.

tools/loop-sandbox/test/sandbox.test.mjs
```diff
@@ -0,0 +1,47 @@
...
+  execSync('git init -b main', { cwd: dir });
+  execSync('git config user.name "Test"', { cwd: dir });
+  execSync('git config user.email "test@example.com"', { cwd: dir });
+  await writeFile(path.join(dir, 'README.md'), '# Test repo');
+  execSync('git add README.md', { cwd: dir });
+  execSync('git commit -m "init"', { cwd: dir });
+  return dir;
+}
+
+test('sandbox run captures patch and cleans up', async () => {
```
cobusgreyling, line 22: **[suggestion]** Only one happy-path test exists. No coverage for: (a) only-this-worktree cleanup when another active worktree exists, (b) path escape `../../` into main tree, (c) binary files, (d) deletions/renames, (e) gitignored untracked files, (f) patch extract failure must not destroy work, (g) worktree/branch gone after success, (h) non-`main` default branch.

**Suggestion:** Add focused tests for cleanup scoping and extract-failure retention first (they lock in the critical bugs); then binary patch round-trip and "main tree unchanged under relative writes only."

tools/loop-sandbox/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "author": "Cobus Greyling",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/cobusgreyling/loop-engineering.git",
+    "directory": "tools/loop-sandbox"
+  },
+  "homepage": "https://cobusgreyling.github.io/loop-engineering/",
+  "dependencies": {
+    "@cobusgreyling/loop-worktree": "file:../loop-worktree"
```
cobusgreyling: **[suggestion]** Dependency is `"@cobusgreyling/loop-worktree": "file:../loop-worktree"`. That matches some monorepo tools for local dev, but published npm consumers cannot resolve `file:` paths. Package also lacks `publishConfig.access: "public"` present on siblings like `loop-worktree`.

**Suggestion:** For publishable packages, use a semver range on the published `@cobusgreyling/loop-worktree` version (and/or workspace protocol if the monorepo adopts it), add `publishConfig`, and document monorepo install order (`build loop-worktree` before `loop-sandbox`).

tools/loop-sandbox/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+# Example output:
+# === Loop Sandbox Patches ===
+# 📄 sandbox-82f1e394.patch (1.2 KB)
+#    Apply: git apply .loop-sandbox/patches/sandbox-82f1e394.patch
+```
+
+## How it works
+
+Under the hood, `loop-sandbox` leverages `@cobusgreyling/loop-worktree` and native `git worktree` primitives. It creates a temporary branch, spawns your process with its `cwd` set to the isolated tree, runs `git diff --cached` to generate the patch, and uses `git worktree remove --force` to clean up.
```
cobusgreyling: **[suggestion]** Safety-sensitive tool does not mention `.gitignore` for `.loop-sandbox/` / `.loop-worktrees/`, human review gates, or `docs/safety.md`. Patches under `.loop-sandbox/patches/` may be committed accidentally; worktree debris may pollute status.

**Suggestion:** Document required gitignore entries, review-before-apply workflow, and link safety/denylist guidance. Consider writing a local `.loop-sandbox/README` or example ignore snippet.

tools/loop-sandbox/src/cli.ts
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env node
+
+import { parseArgs } from 'node:util';
```
cobusgreyling: **[nit]** `parseArgs` is imported from `node:util` but never used. Dead import; `dist/cli.js` does not include it, suggesting src/dist may have drifted or tsc elided inconsistently depending on options.

**Suggestion:** Remove the unused import; ensure `npm run build` regenerates `dist/` so src and dist stay in sync.

tools/loop-sandbox/src/sandbox.ts
```diff
@@ -0,0 +1,145 @@
+import { spawn, execFile } from 'node:child_process';
+import { promisify } from 'node:util';
+import { mkdir, writeFile, readdir, readFile } from 'node:fs/promises';
```
cobusgreyling: **[nit]** `readFile` is imported but unused; `listPatches` dynamically re-imports `node:fs/promises` for `stat` instead of importing `stat` at the top.

**Suggestion:** Drop `readFile`; add `stat` to the static import and use it directly.

tools/loop-sandbox/src/cli.ts
```diff
@@ -0,0 +1,105 @@
...
+        console.log(`To review the changes, run:`);
+        console.log(`  npx @cobusgreyling/loop-sandbox review`);
+        console.log(`\nTo apply the changes immediately, run:`);
+        console.log(`  git apply ${path.relative(root, result.patchFile)}`);
+        console.log('==================================================\n');
+      } else {
+        console.log(`\nℹ️ Run completed with exit code ${result.exitCode}. No changes detected.`);
+      }
+      
+      process.exit(result.exitCode ?? 1);
```
cobusgreyling: **[nit]** On success with no changes, CLI still exits with `result.exitCode ?? 1`. If the child exits 0 with no diffs this is fine; if `exitCode` is unexpectedly `null` (spawn edge cases), the CLI reports failure even though cleanup finished. Minor UX inconsistency with the "No changes detected" message.

**Suggestion:** Use `process.exit(result.exitCode ?? 0)` when the sandbox machinery succeeded, or distinguish child failure vs sandbox infrastructure failure with separate codes.

cobusgreyling (changes requested):
## Summary

`loop-sandbox` is a useful first cut at worktree-isolated agent runs with patch capture, and monorepo wiring is mostly consistent with sibling tools. However, several correctness and safety bugs make it unsafe to treat as an "air-gap" or production L2 guardrail yet.

**Please address the inline findings before merge** (especially cleanup scoping, extract-failure retention, isolation wording, and binary patch fidelity).

### Issue counts by severity
- bugs: 9
- suggestions: 4
- nits: 3

### Loop convention (L1 before L2)
Treat this as **report-only / L1** until isolation claims match the implementation and destructive cleanup paths are fixed. Soften "air-gap" language, link `docs/safety.md`, and add tests for cleanup scoping + extract-failure retention first. Do not present as an unattended L2 safety boundary until those land.

Related: CONTRIBUTING safety checklist + `docs/safety.md`.

THRISHAL12345:
@cobusgreyling 

I have made the changes, please verify them.

cobusgreyling (approved):
## Re-review (after d1f562b)

Author follow-up addresses the CHANGES_REQUESTED findings. Re-checked against the original bug list:

| Finding | Status |
|---|---|
| Blanket `cleanupWorktrees({ statuses: ['active'], force: true })` | Fixed — removes only this worktree path + `branch -D loop/<runId>`, then `gc` |
| Extract failure silent data loss | Fixed — `extractionFailed` skips force cleanup; recovery path logged |
| Overstated air-gap / isolation | Fixed — README WARNING + `docs/safety.md` link |
| Binary patch UTF-8 corruption | Fixed — `gitRaw` with `encoding: 'buffer'`, write Buffer as-is |
| `stdout.trim()` on patch content | Fixed for patch path |
| Hard-coded `main` base | Fixed — default current HEAD + `--base` |
| Orphan `loop/<runId>` branches | Fixed — `git branch -D` after remove |
| `shell: true` by default | Fixed — default false + explicit `--shell` |
| No signal / try-finally cleanup | Fixed — SIGINT/SIGTERM + finally |
| Publishable dep / publishConfig | Fixed — `@cobusgreyling/loop-worktree@^1.2.0` + `publishConfig.access` |
| Tests | 4/4 pass locally: happy path, scoped cleanup, extract-failure retention, binary round-trip |
| Root `test:tools` / `build:tools` | Wired |
| `loop-worktree` main/types exports | Correct for consumers |

**Non-blocking follow-ups (can land later):**
- Soften package.json description (“ultimate L2 safety mechanism”) to match README L1 framing
- Optional `release-loop-sandbox.yml` when ready to publish 1.0.0
- Prefer `markWorktree` + scoped `cleanupWorktrees` over raw git for full manifest hygiene (current path works via `gc` drop)

CI green. Approving and merging.

PR #382, opened 2026-07-25
fix(loop-audit,loop-sandbox): correct gate.yaml auto-fix schema and Windows spawn ENOENT

KhaiTrang1995:
## Summary

Two independent, real bugs found while surveying recent upstream work
(loop-sandbox, MCP loop-gate/loop-audit integration, `--auto-fix`).

## What

**1. `loop-audit --auto-fix` generated a `gate.yaml` that `loop-gate` rejects.**

`loop-gate check` requires `{ version: 1, denylist: string[], maxFiles?,
autoMergeAllowlist? }` and throws on anything else. The auto-fix template
instead wrote `{ gates: [{ name, description, required_approvers }] }` -- a
shape `loop-gate` has never understood. Since `loop-audit`'s own gateYaml
signal only checks that the file exists (not its contents), this passed
audit cleanly while silently breaking `loop-gate check` downstream. Fixed
`GATE_YAML_TEMPLATE` in `tools/loop-audit/src/autofixer.ts` to emit the real
schema, mirroring `templates/gate.yaml.template`.

**2. `loop-sandbox run` fails on Windows for any npm-installed CLI.**

`loop-sandbox` spawns the user's command with `shell: false` by default. On
Windows, npm `.cmd`/`.bat` shims (`npx`, `tsc`, `eslint`, ...) can't be
exec'd directly without a shell and fail with `ENOENT` -- the package's own
README example (`loop-sandbox run -- npx my-agent`) doesn't work as written.
Fixed by retrying once with `shell: true` specifically on `ENOENT` +
Windows, rather than forcing `shell: true` unconditionally (which breaks
commands relying on exact argv quoting, e.g. `node -e "..."`, since cmd.exe
reparses the command line differently).

## Tested

Both packages: full clean rebuild (`rm -rf node_modules dist && npm
install`) + `npm test`, all green.

Added a regression test for the gate.yaml fix (`tools/loop-audit/test/autofixer.test.mjs`)
asserting the generated file actually parses to the schema `loop-gate`
expects.

The sandbox fix went through two more rounds after the first version looked
done: the initial "always use shell:true on Windows" attempt broke an
existing test (`node -e "..."` with embedded quotes), so it was narrowed to
retry-on-ENOENT only. Then, testing the retry path directly against a real
Windows machine turned up a second bug: the failed first attempt's `close`
event still fires a few milliseconds after its `error` event, and without a
guard it would resolve the run's promise with the failed attempt's bogus
exit code *before* the retry (still running) produced its real result --
meaning cleanup could run while the actual command was still executing
against the worktree. Fixed with a small flag that ignores the superseded
first attempt's late events once a retry is in flight, and reproduced the
race locally both before and after the fix to confirm it's actually closed.

### DIFF
--- a/tools/loop-audit/dist/autofixer.js
+++ b/tools/loop-audit/dist/autofixer.js
@@ -147,13 +147,25 @@ const AGENTS_MD_TEMPLATE = `# AGENTS.md — Project Conventions
 - Never auto-merge changes.
 - Ensure all loops isolate their work in worktrees.
 `;
-const GATE_YAML_TEMPLATE = `# gate.yaml - Explicit human approval gates
+const GATE_YAML_TEMPLATE = `# Machine-readable twin of docs/safety.md, enforced by \`loop-gate check\`.
+# See templates/gate.yaml.template in loop-engineering for the full reference.
+version: 1
 
-gates:
-  - name: "Auto-merge to main"
-    description: "Never auto-merge changes to the main branch."
-    required_approvers:
-      - "human"
+denylist:
+  - ".env"
+  - ".env.*"
+  - "**/secrets/**"
+  - "**/credentials/**"
+  - "**/*_key*"
+  - "**/*_secret*"
+
+# Escalate instead of auto-merging when a change touches more than this many files.
+maxFiles: 10
+
+# --action auto-merge only proceeds if every changed path matches one of these.
+autoMergeAllowlist:
+  - "docs/**"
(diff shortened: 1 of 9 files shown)


cobusgreyling (approved):
LGTM — real bugs: gate.yaml auto-fix schema now matches loop-gate; Windows ENOENT retry for npm shims with superseded-close race guarded. Regression test added. CI green.

PR #399, opened 2026-07-26
feat(loop-sandbox): add opt-in advisory lock for scheduled-loop safety

KhaiTrang1995:
## Summary

`loop-sandbox run` isolates a one-shot agent command in a throwaway worktree,
but nothing stopped it from racing a scheduled loop editing the same files
at the same time. `docs/multi-loop.md` already documents a `loop-worktree
lock`/`unlock` convention other loops follow before touching a worktree --
`loop-sandbox` just never used it.

## What

Adds `--lock-paths <globs>` (plus `--lock-owner`, `--lock-ttl`, `--lock-wait`)
to `loop-sandbox run`. When passed, it acquires a `loop-worktree` advisory
lock before creating the worktree and releases it once the run finishes --
including on Ctrl+C mid-run. Off by default; an unadorned `run` is
unprotected, same as before.

`lockPaths`/`unlockOwner` are imported from `loop-worktree`'s `dist/lock.js`
directly rather than its main entry. A first attempt re-exporting them from
`worktree.ts` (the main entry) created a real circular import --
`lock.ts` already imports `MANIFEST_DIR` from `worktree.ts`, so having
`worktree.ts` also re-export from `lock.ts` means neither can finish loading
first, and `MANIFEST_DIR` throws a temporal-dead-zone `ReferenceError` at
import time. `loop-worktree`'s `package.json` has no `exports` field
restricting subpaths, so importing the compiled file directly resolves
cleanly at both runtime and via TypeScript's NodeNext resolution.

Cross-linked from `docs/multi-loop.md` and `tools/loop-sandbox/README.md`'s
new "Multi-loop safety" section.

## Tested

Full clean rebuild (`rm -rf node_modules dist && npm install`) + `npm test`
in both `loop-sandbox` and `loop-worktree`, all green, re-run several times
to rule out flakiness.

The first version of this had a real bug: Ctrl+C during a run only ran the
pre-existing worktree cleanup, never released the lock -- since locks have
no TTL unless `--lock-ttl` is passed, an interrupted run would strand the
lock forever, blocking every future run on those paths until someone ran
`loop-worktree unlock` by hand. Fixed by merging cleanup and lock release
into a single function keyed off state flags (whether a worktree exists,
whether a lock was acquired), and registering the signal handlers before the
lock is even requested rather than after the worktree is created. Confirmed
by manually starting a locked run, killing it mid-flight, and checking the
lock file disappears -- then added a regression test that self-triggers
`process.emit('SIGINT')` inside a dedicated child process (sending a real OS
signal from a test turned out to be unreliable to script on Windows, but
`process.emit` exercises the exact same registered listener a real signal
would).

Also added tests for: the lock being held and released around a normal run,
and a run being rejected before any worktree is created when another owner
already holds an overlapping lock.

### DIFF
--- a/tools/loop-sandbox/README.md
+++ b/tools/loop-sandbox/README.md
@@ -49,8 +49,12 @@ loop-sandbox <command> [options]
 
 | Option | Description |
 |--------|-------------|
-| `--shell` | Forces `shell: true` (for `bash -c`, etc.). On Windows, npm-installed `.cmd` shims (`npx`, `tsc`, ...) that fail with `ENOENT` are automatically retried through a shell, so this is rarely needed there. |
+| `--shell` | Spawns the command using `shell: true` (for bash -c, etc.) |
 | `--base <branch>` | The base branch for the worktree (defaults to current HEAD) |
+| `--lock-paths <globs>` | Comma-separated globs to hold a [`loop-worktree`](../loop-worktree) advisory lock on for the run's duration, so a scheduled loop can't touch the same paths concurrently. Off by default -- see [Multi-loop safety](#multi-loop-safety) below. |
+| `--lock-owner <name>` | Lock owner name (defaults to the run's generated id) |
+| `--lock-ttl <dur>` | e.g. `30m` -- passed through to `loop-worktree`'s `--ttl` |
+| `--lock-wait <dur>` | e.g. `5m` -- passed through to `loop-worktree`'s `--wait` |
 
 ### Examples
 
@@ -63,6 +67,11 @@ npx @cobusgreyling/loop-sandbox run -- npx my-agent
 **Running a shell command:**
 ```bash
 npx @cobusgreyling/loop-sandbox run --shell -- bash -c "echo 'hello' > test.txt"
+```
+
+**Running alongside a scheduled loop that touches the same files:**
+```bash
+npx @cobusgreyling/loop-sandbox run --lock-paths "src/**,docs/**" -- npx my-agent
 ```
 
 **Reviewing and applying patches:**
@@ -79,3 +88,13 @@ npx @cobusgreyling/loop-sandbox review
 ## How it works
 
 Under the hood, `loop-sandbox` leverages `@cobusgreyling/loop-worktree` and native `git worktree` primitives. It creates a temporary branch from your current HEAD, spawns your process with its `cwd` set to the isolated tree, runs `git diff --cached` to generate the patch, and uses `git worktree remove --force` to clean up.
+
+## Multi-loop safety
+
+A sandbox run is not, by itself, protected against a scheduled loop editing
+the same files at the same time -- `--lock-paths` opts a run into
+`loop-worktree`'s advisory lock (acquired before the worktree is created,
+released once the run finishes) so a colliding scheduled loop's own `lock`
+call fails loudly instead of racing silently. See
+[docs/multi-loop.md](../../docs/multi-loop.md) for the wider convention this
+follows.
--- a/tools/loop-sandbox/README.md
+++ b/tools/loop-sandbox/README.md
@@ -49,7 +49,7 @@ loop-sandbox <command> [options]
 
 | Option | Description |
 |--------|-------------|
-| `--shell` | Spawns the command using `shell: true` (for bash -c, etc.) |
+| `--shell` | Forces `shell: true` (for `bash -c`, etc.). On Windows, npm-installed `.cmd` shims (`npx`, `tsc`, ...) that fail with `ENOENT` are automatically retried through a shell, so this is rarely needed there. |
 | `--base <branch>` | The base branch for the worktree (defaults to current HEAD) |
 | `--lock-paths <globs>` | Comma-separated globs to hold a [`loop-worktree`](../loop-worktree) advisory lock on for the run's duration, so a scheduled loop can't touch the same paths concurrently. Off by default -- see [Multi-loop safety](#multi-loop-safety) below. |
 | `--lock-owner <name>` | Lock owner name (defaults to the run's generated id) |
--- a/tools/loop-sandbox/src/sandbox.ts
+++ b/tools/loop-sandbox/src/sandbox.ts
@@ -4,12 +4,29 @@ import { mkdir, writeFile, readdir, stat } from 'node:fs/promises';
 import path from 'node:path';
 import crypto from 'node:crypto';
 import { createWorktree, isGitRepo, gc } from '@cobusgreyling/loop-worktree';
+// loop-worktree's package.json has no "exports" map restricting subpaths, so
+// this resolves cleanly; a re-export from the package's main entry was tried
+// first but creates a module-load cycle (lock.js reads MANIFEST_DIR from
+// worktree.js, which would need lock.js's own export to finish first).
+import { lockPaths, unlockOwner } from '@cobusgreyling/loop-worktree/dist/lock.js';
 
 const runExec = promisify(execFile);
 
 export interface SandboxOptions {
   shell?: boolean;
   base?: string;
+  /**
+   * Glob(s) to hold an advisory loop-worktree lock on for the run's duration,
+   * so a scheduled loop can't touch the same paths concurrently. Unset by
+   * default -- sandbox runs are unprotected unless the caller opts in.
+   */
+  lockPaths?: string[];
+  /** Lock owner name; defaults to the run's generated id. */
+  lockOwner?: string;
+  /** e.g. "30m" -- passed straight through to loop-worktree's --ttl. */
+  lockTtl?: string;
+  /** e.g. "5m" -- passed straight through to loop-worktree's --wait. */
+  lockWait?: string;
 }
 
 export interface SandboxResult {
@@ -45,7 +62,7 @@ export async function runInSandbox(root: string, command: string, args: string[]
   }
 
   const runId = `sandbox-${crypto.randomBytes(4).toString('hex')}`;
-  
+
   // 1. Setup paths
   const sandboxDir = path.join(root, '.loop-sandbox');
   const patchesDir = path.join(sandboxDir, 'patches');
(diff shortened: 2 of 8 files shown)


tools/loop-sandbox/src/sandbox.ts
```diff
@@ -45,52 +62,51 @@ export async function runInSandbox(root: string, command: string, args: string[]
...
-  // Signal handlers for graceful cleanup if user ctrl+c's during run
+  // Cleanup/lock-release must run on every exit path, including Ctrl+C mid-run
+  // -- both are guarded on state (lockHeld / worktreeAbsPath) rather than
+  // living in separate handlers, since a signal can land before, during, or
+  // after the worktree exists, and an unreleased lock (no --lock-ttl means it
+  // never expires on its own) would otherwise strand every future lock on the
+  // same paths until someone runs `loop-worktree unlock` by hand.
   let isCleaningUp = false;
   const cleanup = async () => {
     if (isCleaningUp) return;
```
cobusgreyling: **[bug]** `cleanup` uses a boolean `isCleaningUp` guard that makes a second call a no-op, while `sigHandler` always does `cleanup().finally(() => process.exit(1))`. A second Ctrl+C/SIGTERM (very common when `git worktree remove` is slow) therefore schedules an immediate `process.exit(1)` without waiting for the in-flight cleanup. Unlock runs only after worktree teardown (lines 89–110), so the process can die mid-`worktree remove` with the advisory lock still on disk. With no default TTL, that strands every future overlapping `lock` until a manual `loop-worktree unlock`.

**Suggestion:** Keep a single shared `cleanupPromise` and have every signal handler await that same promise before exiting (never exit on a no-op second entry). Prefer releasing the lock first (or in parallel with) worktree teardown so the critical resource is freed even if later steps hang or a forced exit lands. Optionally stop re-arming exit on subsequent signals once cleanup has started.

tools/loop-sandbox/src/sandbox.ts
```diff
@@ -45,52 +62,51 @@ export async function runInSandbox(root: string, command: string, args: string[]
...
 
-    console.log(`🧹 Cleaning up sandbox worktree...`);
-    try {
-      await git(['worktree', 'remove', '--force', worktreeAbsPath], root);
-      await git(['branch', '-D', `loop/${runId}`], root).catch(() => {});
-      await gc({ root, force: false });
-    } catch (err) {
-      console.error(`❌ Failed to cleanup sandbox worktree. It may need manual removal:`, err);
-      process.exitCode = 1;
+    if (lockHeld) {
```
cobusgreyling: **[suggestion]** Unlock is gated on `lockHeld`, which is only set after `lockPaths(...)` resolves successfully (line 124). During `--lock-wait`, `loop-worktree` writes a `${owner}.wait.json` while polling; if the user interrupts mid-wait, cleanup skips `unlockOwner`, so the wait file is left until its wait expiry (and can participate in deadlock detection until then). `unlockOwner` already clears both lock and wait files for that owner.

**Suggestion:** Track “we attempted lock under this owner” (or always best-effort `unlockOwner(root, lockOwner)` when `options.lockPaths` was set), so SIGINT during wait cleans the wait file the same way a completed/failed wait path would.

tools/loop-sandbox/src/sandbox.ts
```diff
@@ -4,12 +4,29 @@ import { mkdir, writeFile, readdir, stat } from 'node:fs/promises';
 import path from 'node:path';
 import crypto from 'node:crypto';
 import { createWorktree, isGitRepo, gc } from '@cobusgreyling/loop-worktree';
+// loop-worktree's package.json has no "exports" map restricting subpaths, so
+// this resolves cleanly; a re-export from the package's main entry was tried
+// first but creates a module-load cycle (lock.js reads MANIFEST_DIR from
+// worktree.js, which would need lock.js's own export to finish first).
+import { lockPaths, unlockOwner } from '@cobusgreyling/loop-worktree/dist/lock.js';
```
cobusgreyling, line 16: **[suggestion]** Importing `@cobusgreyling/loop-worktree/dist/lock.js` reaches a private compiled path. It works today because `loop-worktree` has no `exports` map and publishes `dist/`, and the PR correctly explains the cycle if re-exported from `worktree.ts`. This still couples `loop-sandbox` to package layout: any future `exports` lockdown, dist rename, or dual package hazard breaks consumers at runtime.

**Suggestion:** Prefer a supported public surface in a follow-up (e.g. `lock.js` as a package subpath export, or a tiny `index` that re-exports worktree + lock without a cycle by moving `MANIFEST_DIR` to a shared constants module). Until then, pin/document the dependency on this path in the README multi-loop section so a worktree packaging change is treated as breaking for sandbox.

tools/loop-sandbox/src/sandbox.ts
```diff
@@ -114,7 +154,7 @@ export async function runInSandbox(root: string, command: string, args: string[]
 
         const attempt = (useShell: boolean) => {
           const child = spawn(command, args, {
```
cobusgreyling, line 179: **[suggestion]** On SIGINT, cleanup removes the worktree and releases the lock while the sandboxed child (spawned with `stdio: 'inherit'`, never tracked for kill) may still be alive. After unlock, a scheduled loop can acquire the same paths while an orphaned agent continues writing. Real terminal Ctrl+C often signals the whole process group, but programmatic signals / some environments do not; the new SIGINT test even documents cleanup while the child is still running.

**Suggestion:** Keep a reference to the active child and `child.kill(...)` (and await `close`) before unlocking, so advisory protection is not dropped under a still-running agent. This matters more now that unlock is part of the signal path.

tools/loop-sandbox/README.md
```diff
@@ -49,8 +49,12 @@ loop-sandbox <command> [options]
 
 | Option | Description |
 |--------|-------------|
-| `--shell` | Forces `shell: true` (for `bash -c`, etc.). On Windows, npm-installed `.cmd` shims (`npx`, `tsc`, ...) that fail with `ENOENT` are automatically retried through a shell, so this is rarely needed there. |
+| `--shell` | Spawns the command using `shell: true` (for bash -c, etc.) |
```
cobusgreyling: **[nit]** While adding lock options, the `--shell` row lost the Windows ENOENT / `.cmd` shim retry guidance that still exists in CLI help and in `sandbox.ts`. That is a user-facing docs regression unrelated to the feature.

**Suggestion:** Restore the Windows retry note on the `--shell` row (or point at the CLI help text) so the table stays accurate.

tools/loop-sandbox/test/sandbox.test.mjs
```diff
@@ -110,3 +115,83 @@ test('binary patch captures exactly without utf8 corruption', async () => {
     await rm(root, { recursive: true, force: true }).catch(() => {});
   }
 });
+
```
cobusgreyling, line 118: **[nit]** Coverage checks post-run release, overlap rejection, and single synthetic SIGINT, but not (a) that the lock is actually held mid-run, (b) extraction-failure still unlocks, or (c) double-signal / in-flight cleanup. The extraction-failure unlock fix is important and untested.

**Suggestion:** Add a short mid-run assertion (e.g. agent script calls out / writes after `listLocks` sees the owner) and an extraction-failure + `lockPaths` case asserting `listLocks` is empty while the worktree remains. Double-SIGINT can wait on the shared-promise fix.

cobusgreyling (changes requested):
## Summary

Solid, well-scoped opt-in for multi-loop advisory locking: acquire before worktree create, release on every exit path (including the extraction-failure path that previously early-returned), and tests cover happy-path release, overlapping refusal, and single SIGINT. The main correctness gap is cleanup/signal interaction: a second SIGINT/SIGTERM while cleanup is in the slow worktree-remove phase can `process.exit` before unlock runs, stranding a no-TTL lock. Minor docs and packaging nits aside, this is close; fix the double-signal strand path before merge.

## Issue counts by severity

- bugs: 1
- suggestions: 3
- nits: 2

---
**Decision: request changes.** Please fix the double-SIGINT/SIGTERM cleanup path so unlock cannot be skipped mid-worktree-remove (shared cleanup promise / unlock-first). Other items can land as follow-ups.

KhaiTrang1995:
- **Critical bug (double-signal exit):** confirmed. The old `cleanup()` used a boolean guard, so a second SIGINT/SIGTERM during an in-flight cleanup returned immediately and its `.finally(() => process.exit(1))` could fire before the *first* signal's lock release and worktree teardown had finished. Fixed by sharing one `cleanupPromise` across every signal handler call, so every invocation (first or repeated) awaits the same in-flight cleanup before exiting. To make sure the regression test actually catches this rather than passing by coincidence, I reverted to the old guard locally, reran the test, and confirmed it left a stale `sandbox-*` worktree directory behind — then confirmed the fix removes it cleanly.
- **Wait-file cleanup:** lock release is now gated on "a lock was requested" rather than "a lock was confirmed acquired," so an interrupted `--lock-wait` still gets its `<owner>.wait.json` cleaned up. `unlockOwner` is a safe no-op if neither file exists, so this is unconditional.
- **Child process lifecycle:** the sandboxed command is now killed before the lock is released or the worktree is torn down, instead of being left running underneath cleanup.
- **Private `dist/lock.js` import:** left as-is for this PR to keep the diff scoped, but added a comment noting it's a known internal-API dependency and that a public `loop-worktree/lock` subpath export would be a good follow-up.
- **Docs regression:** restored the `--shell` row's note about Windows `.cmd`-shim ENOENT auto-retry.
- **Test coverage:** added tests for the lock being held mid-run (not just after), the lock still being released when patch extraction fails, and the double-signal case above.

cobusgreyling (approved):
## Re-review (post-fix commit b5bef1f)

Thanks for the thorough follow-up. The shared `cleanupPromise` correctly closes the double-SIGINT strand window: every signal awaits the same in-flight cleanup before `process.exit`, unlock runs before worktree teardown, and the child is killed first. Gating unlock on lock *requested* (not only acquired) also covers interrupted `--lock-wait` wait-files.

### Verified
- Double-signal + mid-run lock + extract-fail release tests look solid and target the real failure modes
- Docs restore of the `--shell` Windows `.cmd`-shim note
- Comment documenting the private `dist/lock.js` import + public-export follow-up

### Follow-ups (non-blocking)
- Public `loop-worktree/lock` subpath export so consumers don't reach into `dist/`
- Optional default TTL when `--lock-paths` is set without `--lock-ttl` (defense in depth against any future strand path)

**Decision: approve.** Ready to merge.

cobusgreyling:
Merged — thank you @KhaiTrang1995 for the careful lock integration and for addressing the double-SIGINT cleanup path with solid regression tests. Follow-up welcome: public `loop-worktree/lock` export.

PR #404, opened 2026-07-27
Harden loop-action command execution against unquoted shell expansion

cobusgreyling:
## Summary

Same change as #395 (from @amarjaleelbanbhan), re-hosted on a same-repo branch so required `audit` / `validate` checks can run (fork PR workflows were stuck in `action_required`).

`inputs.command` was spliced as raw text into the `run:` script before bash parsed it. For multi-line `command: |` blocks this silently broke sandbox isolation: only the first line reached `loop-sandbox run -- ...`; every later line executed as a bare top-level statement on the runner, bypassing the sandbox and circuit-breaker entirely.

Write `command` to a temp script (passed via `env:`, never re-templated by the workflow YAML) and invoke it as a single quoted `bash -eo pipefail "$SCRIPT"` call from both the sandbox and direct paths. Document safe usage and warn against embedding untrusted input in `command` in the README.

Closes #395 (supersedes / merges the same commits).

## Attribution

- Original PR: https://github.com/cobusgreyling/loop-engineering/pull/395
- Author: @amarjaleelbanbhan

## Checklist

- [x] Security-sensitive command handling documented
- [x] Fail-fast (`bash -eo pipefail`) on all execution paths


cobusgreyling:
Superseded by merge of **#395** (same commits; original author @amarjaleelbanbhan).

This same-repo branch was only opened so required checks could run while the fork PR’s workflows sat in `action_required`. Cleaning up.

PR #430, opened 2026-07-29
fix(loop-worktree): prevent manifest entry loss during concurrent ope…

AIMindCrafter:
…rations (#408)

## Changes
- [ ] New pattern or starter (followed `templates/pattern-template.md` + updated `registry.yaml`)
- [ ] Doc / example improvement
- [ ] Tool change (loop-audit)
- [ ] Story (includes real failure or surprise + lesson)

## Checklist (from CONTRIBUTING)
- [ ] All required sections present for patterns
- [ ] Links work from README, patterns/README, starters/README, docs/index
- [ ] No secrets, tokens, internal company URLs
- [ ] `STATE.md*` examples use `.example` suffix
- [ ] Safety-related content references `docs/safety.md`
- [ ] Ran `node tools/loop-audit/dist/cli.js .` (or on the starter) and addressed findings

## Testing / Dogfood
- [ ] `loop-audit` passes on affected starters or this repo
- [ ] Manual review of generated state / skill output

## Screenshots / Examples (if UI or command output)

---

*This template enforces the high bar this reference is known for.*

PR #407, opened 2026-07-28
feat(loop-worktree): expose public lock API

shixi-li:
## Summary

Expose the advisory lock coordinator through the supported
`@cobusgreyling/loop-worktree/lock` subpath, following the explicit follow-up
welcomed in #399.

## Why

`loop-sandbox` currently reaches into
`@cobusgreyling/loop-worktree/dist/lock.js`. That private build path makes
consumers depend on package layout rather than a public contract. This PR
establishes the package contract without changing lock runtime behavior.

The `loop-sandbox` import migration is intentionally deferred until
`loop-worktree@1.3.0` is published and #398 has settled, avoiding a
cross-package release dependency and overlapping changes.

## Changes

- add root and `./lock` conditional exports with declaration mappings
- retain `./dist/*` and `./package.json` compatibility exports
- add `typesVersions` support for classic TypeScript module resolution
- bump `loop-worktree` to 1.3.0, align the previously stale lockfile, and
  update the publish checklist
- document the public library imports
- run the existing lock suite through the public package self-reference
- add a packed, clean-consumer test covering root, public lock, legacy deep
  imports, and NodeNext/classic TypeScript resolution
- invoke npm portably in the consumer test and keep cleanup compatible with
  the full declared Node 18 range

## Scope

This PR does not modify `loop-sandbox`, manifest concurrency, or the open #398
work.

## Validation

- `cd tools/loop-worktree && npm test` — 32/32 passed
- `cd tools/loop-audit && npm run build && node dist/cli.js ../..` — 100/100 L3
- `git diff --check`

Follow-up to #399.

PR #433, opened 2026-07-29
docs: document public loop-worktree lock import in QUICKSTART (#426)

AIMindCrafter:
## Changes
- [ ] New pattern or starter (followed `templates/pattern-template.md` + updated `registry.yaml`)
- [ ] Doc / example improvement
- [ ] Tool change (loop-audit)
- [ ] Story (includes real failure or surprise + lesson)

## Checklist (from CONTRIBUTING)
- [ ] All required sections present for patterns
- [ ] Links work from README, patterns/README, starters/README, docs/index
- [ ] No secrets, tokens, internal company URLs
- [ ] `STATE.md*` examples use `.example` suffix
- [ ] Safety-related content references `docs/safety.md`
- [ ] Ran `node tools/loop-audit/dist/cli.js .` (or on the starter) and addressed findings

## Testing / Dogfood
- [ ] `loop-audit` passes on affected starters or this repo
- [ ] Manual review of generated state / skill output

## Screenshots / Examples (if UI or command output)

---

*This template enforces the high bar this reference is known for.*

PR #440, opened 2026-07-30
fix(release): bump loop-worktree to 1.3.1 and correct the publish runbook

shixi-li:
## Summary

Unblock the stalled 2026-07-29 npm publish batch: bump `loop-worktree` 1.3.0 → 1.3.1 so the release can ride a fresh tag, and correct `RELEASE_NOTES_DRAFT.md`'s package status + publish sequence to match what the 07-29 runs actually showed — including the registry-side step behind every `E404 PUT`.

## Root cause

Two independent blockers took down the 07-29 batch (3 dispatch rounds, 15 failed runs):

1. **`loop-worktree-v1.3.0` is a burned tag.** It points at `1c88a5e` (#422), which predates #423's pack-tarball directory-scan fix. Release dispatches check out the tag, so every run died in `package-exports.test.mjs` (`ERR_INVALID_ARG_TYPE`, 31/32) under the workflow's `npm@latest` before reaching publish — no further `main`-side fixing can help that tag. npm never received 1.3.0.
2. **Every other package failed at the registry, not in CI.** audit/init/mcp-server/sandbox cleared tests/build and failed at `npm publish` with `404 Not Found - PUT` — including the 15:10 round, which already ran token-less after #422. In each log the Sigstore provenance uploads *before* the PUT fails, so the GitHub OIDC token itself works; the registry is rejecting an **unauthenticated** publish (npm masks unauthorized as 404). That points at the npmjs.com-side trusted-publisher relationship (configured per package, exact workflow filename), which no workflow change can fix.

## Changes

- bump `tools/loop-worktree` to **1.3.1** (package.json + lockfile). Semver-patch: since the burned tag, loop-worktree gained only #423 (test hardening) and #430 (concurrency manifest-loss fix)
- `RELEASE_NOTES_DRAFT.md`: correct the worktree highlight (1.3.1 + why, credit #430), update the package-status table to post-07-29 reality, document the E404 root cause, and rewrite the publish sequence — trusted-publisher config → tag `loop-worktree-v1.3.1` → re-dispatch the existing green tags → first-publish path for sandbox/swarm
- `scripts/issue-bodies/quickstart-worktree-lock-import.md`: 1.3.0 → 1.3.1 (1.3.0 will never exist on npm)

## Scope

No runtime code changes. `tools/loop-sandbox`'s `^1.2.0` range already resolves 1.3.1; migrating its deep `dist/lock.js` import to the public subpath stays deferred (as agreed in #407's review) until 1.3.1 is actually on npm.

## Validation

- `cd tools/loop-worktree && npm test` — 35/35 (npm 11.13.0, node 24)
- `cd tools/loop-worktree && npx -y npm@latest test` — 35/35 (npm 12.0.2, matching the release workflow's `npm install -g npm@latest`)
- negative control: `git worktree` at `1c88a5e` (current `loop-worktree-v1.3.0` target), same npm 12.0.2 → 31/32 with the exact `ERR_INVALID_ARG_TYPE` from run 30464568551
- `git diff --check` — clean

Follow-up to #407 / #421–#423 / #430.

### DIFF
--- a/RELEASE_NOTES_DRAFT.md
+++ b/RELEASE_NOTES_DRAFT.md
@@ -1,4 +1,4 @@
-# Release notes draft — week of 2026-07-29 (updated 2026-07-29)
+# Release notes draft — week of 2026-07-29 (updated 2026-07-30)
 
 **Status:** Draft for human review ([#332](https://github.com/cobusgreyling/loop-engineering/issues/332)). Edit before publishing a discussion post or tagging packages.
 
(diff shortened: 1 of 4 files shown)


cobusgreyling (approved):
## Approved

Clear unblock for the burned `loop-worktree-v1.3.0` tag and the 07-29 E404 publish failures.

- Semver-patch 1.3.0 → 1.3.1 is the right move (tag rewrite avoided)
- RELEASE_NOTES_DRAFT correctly documents trusted-publisher + first-publish path
- Issue-body version string updated

Note (non-blocking): `scripts/issue-bodies/quickstart-worktree-lock-import.md` still shows `unlockPaths` — that name does not exist (API is `unlockOwner`). Same typo as #433; fix can land with that docs PR.

Merging so the release can ride a fresh tag.

PR #446, opened 2026-07-31
fix(loop-sandbox): don't claim a lock was released when it was never acquired

KhaiTrang1995:
## Summary

`cleanup()` logged `Releasing lock held by "X"...` unconditionally
whenever a run requested a lock, even when `lockPaths()` itself had
thrown (e.g. blocked by another owner with no `--wait`) and no lock
file was ever written. The message implied a release that never
happened.

Now the message only prints once `unlockOwner()`'s return value
confirms something (a lock, or a stray `--lock-wait` wait file) was
actually removed.

Found this in a self-review pass on `loop-sandbox`'s advisory-lock
integration (#399, merged) while re-checking the cleanup path's
messaging against every exit case, not just the ones already covered
by tests.

## Test plan

- Added a test that captures `console.log` output and asserts no
  release message prints when `lockPaths()` fails outright (blocked
  by another owner).
- Added an assertion to the existing "holds and releases" test
  confirming the release message does print when a lock is actually
  acquired and released.
- Verified the new assertion actually discriminates: reverted
  `sandbox.ts` to the pre-fix version, reran the test, confirmed it
  fails with the exact misleading message
  (`Releasing lock held by "sandbox-test-owner"...`), then restored
  the fix and reran the full suite (10/10 passing).
  of `loop-worktree` and `loop-sandbox` before testing.

### DIFF
--- a/tools/loop-sandbox/dist/sandbox.js
+++ b/tools/loop-sandbox/dist/sandbox.js
@@ -74,12 +74,21 @@ export async function runInSandbox(root, command, args, options = {}) {
                 activeChild?.kill();
                 // Release the lock before worktree teardown: it's the resource other
                 // loops are blocked on, and a slow worktree removal shouldn't delay
-                // it.
+                // it. Logged only once we know a lock (or stray wait file) actually
+                // existed -- lockRequested only means a lock was asked for, so
+                // logging unconditionally here would print "Releasing lock held by
+                // X" even when lockPaths() itself failed (e.g. blocked by another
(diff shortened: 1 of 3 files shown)


cobusgreyling (approved):
Good catch on the misleading release log. Cleanup now only claims a release when `unlockOwner` actually removed something, with regression coverage for both the acquired and blocked paths. LGTM.

cobusgreyling:
Merged. Thanks @KhaiTrang1995 — the release-log fix plus regression tests are exactly the kind of self-review catch we want.

PR #448, opened 2026-08-01
fix(loop-sandbox): use loop-worktree's public lock subpath instead of a dist/ deep import

KhaiTrang1995:
## Summary

loop-worktree 1.3.0 (#407) added a supported `@cobusgreyling/loop-worktree/lock`
export (lockPaths, unlockOwner, LOCKS_DIR) specifically so consumers wouldn't
need to reach into `dist/`. loop-sandbox's advisory-lock integration (#399)
predates that export and was importing from
`@cobusgreyling/loop-worktree/dist/lock.js` directly, with a comment flagging
it as a known workaround pending a public entry point.

Switches the import to the supported `@cobusgreyling/loop-worktree/lock`
subpath and bumps the dependency floor to `^1.3.0` (the first version whose
package.json exports map actually defines `./lock` -- an install pinned
below that has no such subpath to resolve).

## Test plan

  of both `loop-worktree` and `loop-sandbox`.
- `npm test` in `loop-worktree`: 35/35 passing, including the package's own
  `packed package exposes the public lock subpath and keeps legacy deep
  imports` test.
- `npm test` in `loop-sandbox`: 10/10 passing, unchanged behavior since this
  is an import-path swap with no logic changes.

### DIFF
--- a/tools/loop-sandbox/dist/sandbox.js
+++ b/tools/loop-sandbox/dist/sandbox.js
@@ -4,16 +4,7 @@ import { mkdir, writeFile, readdir, stat } from 'node:fs/promises';
 import path from 'node:path';
 import crypto from 'node:crypto';
 import { createWorktree, isGitRepo, gc } from '@cobusgreyling/loop-worktree';
-// loop-worktree's package.json has no "exports" map restricting subpaths, so
-// this resolves cleanly; a re-export from the package's main entry was tried
-// first but creates a module-load cycle (lock.js reads MANIFEST_DIR from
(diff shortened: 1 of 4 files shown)


cobusgreyling (approved):
Correct follow-through on the public `@cobusgreyling/loop-worktree/lock` subpath from 1.3.0. Dependency floor bump to `^1.3.0` is right. LGTM.

cobusgreyling:
Merged. Thanks @KhaiTrang1995 — sandbox now uses the supported `@cobusgreyling/loop-worktree/lock` subpath.

PR #451, opened 2026-08-02
fix(docs): stop Live Telemetry charts from growing endlessly + pin CDN script versions

KhaiTrang1995:
## Summary

The three charts in the Dogfooding / Live Telemetry section
(https://cobusgreyling.github.io/loop-engineering/#telemetry) grew taller
forever, making the page feel like it never finished loading. Root cause:
each canvas sat directly inside a `.card` div with no explicit height.
Chart.js is configured with `responsive: true` + `maintainAspectRatio:
false`, which resizes the canvas to fill its parent -- but since the
container's own height was determined by the canvas inside it, every
resize grew the container, which re-triggered Chart.js's ResizeObserver,
which grew the canvas again. Unbounded feedback loop.

Fixed by wrapping each canvas in a `position: relative; height: 240px`
container, so the container's size no longer depends on the canvas it
holds.

Also, while investigating:
- Both third-party scripts on the page were unpinned (`mermaid@11` floats
  across every 11.x release, and the Chart.js `<script>` tag had no
  version at all) -- any upstream release of either library can silently
  change or break the live page with zero corresponding commit here.
  Pinned both to the versions currently being served (mermaid 11.16.0,
  Chart.js 4.5.1).
- `renderTelemetry()`'s cumulative-tokens calculation recomputed the full
  prefix sum from scratch for every point (O(n^2) over loop-run-log.md's
  entries) instead of carrying a running total. loop-run-log.md is
  appended to daily with nothing pruning it, so this cost compounds
  indefinitely. Replaced with an O(n) running total.

## Test plan

- Reproduced the growth bug with a headless-Chromium probe before fixing:
  `canvas.height` climbed from 2661px to 8152px over 5 seconds with zero
  user interaction, and `body.scrollHeight` grew on every scroll.
- Re-ran the same probe after the fix: `canvas.height` holds steady
  indefinitely, and repeated scrolling no longer changes `scrollHeight`.
- Verified the O(n) token-sum refactor produces identical output to the
  old O(n^2) version on a sample dataset.
- Verified both pinned CDN URLs resolve (200 OK) and the page's inline
  script still passes `node --check` after the edits.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -15,7 +15,7 @@ title: Loop Engineering — Showcase
   <link rel="stylesheet" href="{{ site.baseurl }}/assets/css/showcase.css">
   <link rel="icon" href="https://raw.githubusercontent.com/cobusgreyling/loop-engineering/main/assets/visuals/loop-engineering-logo.svg">
   <script type="module">
-    import mermaid from "https://cdn.jsdelivr.net/npm/mermaid@11/dist/mermaid.esm.min.mjs";
+    import mermaid from "https://cdn.jsdelivr.net/npm/mermaid@11.16.0/dist/mermaid.esm.min.mjs";
     mermaid.initialize({
       startOnLoad: true,
       theme: "dark",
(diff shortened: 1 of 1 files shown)


cobusgreyling (approved):
Solid root-cause fix for the Chart.js resize feedback loop (fixed-height wrapper), plus sensible CDN pins and the O(n) running total. LGTM.

cobusgreyling:
Merged. Thanks @KhaiTrang1995 — Live Telemetry charts no longer unbounded-grow, CDN pins are locked, and the token sum is O(n).

PR #475, opened 2026-08-09
fix(readiness-core): deduplicate scanned skills to avoid double-counting score

pxmpsdev:
Prevents double-counting when the same skill name exists in multiple skill directories (e.g. .grok/skills/foo and skills/foo). Previously scanSkillDirectories pushed duplicates, inflating signals.skills.count and awarding skillsTwoPlus (14 pts) instead of skillsOne (7 pts), skewing the Loop Readiness score.

Verification:
- npm run build && node --check dist/index.js ok
- npm test in tools/readiness-core -> 3/3 pass
- Manual fixture with duplicate skill name in two dirs now correctly deduplicated


cobusgreyling (approved):
Thanks @pxmpsdev — clean, focused fix with clear verification. Merging.

PR #476, opened 2026-08-09
fix(loop-sync): handle CRLF frontmatter correctly

pxmpsdev:
Fix frontmatter parsing to handle Windows line endings (CRLF). Previously `extractFrontmatter` only split on `\n`, leaving `\r` in values and accepting `---hello` as frontmatter. Now requires `---\n` or `---\r\n` and splits on `/\r?\n/`, correctly handling `\r\n` files and rejecting invalid frontmatter without newline.

Verification:
- Manual test with LF and CRLF fixtures – both parse correctly, `---hello` correctly rejected
- Existing LF behaviour preserved (checked via node snippet)
- Minimal change, no breaking API

### DIFF
--- a/tools/loop-sync/src/sync.ts
+++ b/tools/loop-sync/src/sync.ts
@@ -67,13 +67,13 @@ function extractFrontmatter(content: string): { frontmatter: Record<string, stri
   const frontmatter: Record<string, string> = {};
   let body = content;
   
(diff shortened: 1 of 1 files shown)


cobusgreyling (approved):
Thanks @pxmpsdev — clean, focused fix with clear verification. Merging.

PR #518, opened 2026-08-15
fix(loop-worktree): don't crash gc() when a worktree dir was removed out-of-band

KhaiTrang1995:
## Problem
gc()'s gitWorktreePaths() calls realpath() on every path reported by
`git worktree list --porcelain` with no error handling. git keeps
listing a worktree (annotated "prunable") even after its directory is
gone -- a manual rm -rf, a crash mid-cleanup, a container wipe --
rather than removed via `git worktree remove`. realpath() on that
now-missing path throws ENOENT, which propagated out of gc() entirely
instead of surfacing the entry as `dropped`, aborting the exact
reconciliation gc() exists to perform.

## Fix
Treat a missing directory as simply absent from disk instead of
letting the exception escape. The manifest entry (if any) is now
correctly reconciled as `dropped`.

## Test plan
Added a regression test that deletes a worktree directory directly
(bypassing `git worktree remove`) and asserts gc() completes and
reports it as dropped. Full clean rebuild + npm test: 36/36 passing.

### DIFF
--- a/tools/loop-worktree/dist/worktree.js
+++ b/tools/loop-worktree/dist/worktree.js
@@ -215,7 +215,22 @@ async function gitWorktreePaths(root) {
         if (!line.startsWith('worktree '))
             continue;
         const abs = line.slice('worktree '.length).trim();
-        const absReal = await realpath(abs);
+        let absReal;
+        try {
+            absReal = await realpath(abs);
+        }
+        catch (err) {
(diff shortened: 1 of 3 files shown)


cobusgreyling (approved):
Looks correct: ENOENT from realpath() on a prunable/missing worktree should not abort gc(). Matches the existing out-of-band drop behavior and the new regression test covers the crash path.

PR #502, opened 2026-08-11
fix(loop): use pathToFileURL for dynamic import in files.test (Windows)

TayyabHussain03:
## Problem

`npm test` in `tools/loop` crashes on Windows with:

Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]:
Only URLs with a scheme in: file, data, and node are supported.
On Windows, absolute paths must be valid file:// URLs.
Received protocol 'd:'

`path.resolve()` returns a Win32 absolute path (`D:\...`).
Node's ESM loader rejects this for dynamic `import()` calls on Windows.
This caused `test/files.test.mjs` to crash entirely — 0 of 3 tests ran.

## Fix

```diff
- import { fileURLToPath } from 'node:url';
+ import { fileURLToPath, pathToFileURL } from 'node:url';

- const files = await import(path.resolve(__dirname, '../dist/files.js'));
+ const files = await import(pathToFileURL(path.resolve(__dirname, '../dist/files.js')).href);
```

`pathToFileURL` is already available via `node:url` — one import added, one call wrapped.

## Result

| | Before | After |
|---|---|---|
| `files.test.mjs` | crash (0/3 ran) | ✅ 3/3 pass |
| Total `tools/loop` | 5/8 pass | 8/8 pass* |

*2 remaining failures (`readiness-core` not found, `doctor --json`) are
a separate monorepo workspace issue — not related to this fix.

## Environment
- Windows 11
- Node.js v24.14.0
- Repo cloned to `D:\` drive

## Checklist
- [x] Tool change (tools/loop)
- [x] Ran `npm test` in `tools/loop` — 8/8 pass after fix
- [x] No secrets, tokens, or internal company URLs
- [x] Allow edits by maintainers ✓

PR #519, opened 2026-08-15
fix(scripts): stop pruning run-log entries with non-ISO run_id

KhaiTrang1995:
## Problem
scripts/append-run-log.mjs prunes entries older than 30 days using
new Date(obj.run_id).getTime(). JS's Date constructor is lenient: a
non-ISO run_id (a numeric GitHub run id, a custom slug like "run-1")
doesn't reliably parse to NaN -- it can parse into a spurious
in-range-looking date instead ("run-1" -> 2001-01-01), which then
reads as 30+ days old and gets silently deleted on the very next
append, even though the entry was just written. tools/loop-metrics
already handles this exact run_id shape correctly (keep on
unparseable rather than drop); this script did the opposite.

## Fix
Only strings shaped like an ISO date (^\d{4}-\d{2}-\d{2}) are parsed
as timestamps for pruning at all; anything else is always kept,
matching loop-metrics' handling.

Also fixes append-run-log.test.mjs's own SCRIPT path, which used
new URL(...).pathname -- on Windows that yields a leading-slash path
("/D:/...") that node's CLI mis-resolves relative to the current
drive instead of as absolute (the same class of bug already fixed for
tools/loop/test/files.test.mjs in #502). Switched to fileURLToPath().

## Test plan
Added a regression test appending an entry with a non-ISO run_id
("run-1") followed by a second append, asserting the first entry
survives. node --test scripts/append-run-log.test.mjs: 3/3 passing
(this also required the Windows path fix to even run locally).

### DIFF
--- a/scripts/append-run-log.mjs
+++ b/scripts/append-run-log.mjs
@@ -7,6 +7,15 @@ import { readFile, writeFile } from 'node:fs/promises';
 
 const MARKER = '<!-- Loop appends below this line -->';
 const MAX_AGE_MS = 30 * 24 * 60 * 60 * 1000;
+// Only strings shaped like an ISO date are treated as timestamps for pruning.
+// `new Date(...)` is a lenient parser: a non-ISO run_id (a numeric GitHub run
+// id, a custom slug like "run-1") doesn't reliably yield NaN, it can parse
+// into a spurious in-range-looking date instead (e.g. "run-1" -> 2001-01-01),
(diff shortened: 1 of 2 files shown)


cobusgreyling (approved):
Correct fix. JS Date is too lenient on slugs like run-1; gating prune on ISO-shaped run_id matches loop-metrics and the regression test locks it. fileURLToPath() for the test script path is the right Windows fix.

PR #523, opened 2026-08-15
 feat(loop-sync): make --auto-fix actually fix things

KhaiTrang1995:
## Problem
--auto-fix was parsed by the CLI, documented in --help, and threaded
into runSync(), but runSync() never read it back -- the flag was a
complete no-op. Running loop-sync . --auto-fix produced identical
output to running without the flag, with no indication that nothing
had been fixed. --dry-run had the same problem: destructured, never
used.

## Fix
Implements real auto-fix, scoped to what can be fixed correctly
without guessing project-specific intent:

- Missing STATE.md, gate.yaml, loop-budget.md, and loop-run-log.md are
  scaffolded with minimal, pattern-agnostic defaults (loop-run-log.md
  keeps the marker append-run-log.mjs depends on; gate.yaml ships a
  secure-by-default denylist).
- A LOOP.md that exists but doesn't reference STATE.md gets a
  reference line appended.
- LOOP.md and AGENTS.md are never fabricated when missing -- their
  content is pattern-specific (the goal, the workflow), so guessing it
  would be actively wrong. Those still just point at loop-init, now
  with --auto-fix mentioned as an option for the rest.
- --dry-run now does what it always claimed to: with --auto-fix, it
  reports what would be scaffolded/changed without writing anything.

The scaffold content is inlined in source rather than read from this
monorepo's templates/*.template files, since loop-sync ships as a
standalone npm package (files: ["dist"]) that runs against arbitrary
target directories -- a path relative to the installed package
wouldn't resolve once installed outside this checkout.

## Test plan
4 new tests covering scaffold-on-missing, LOOP.md/AGENTS.md never
being fabricated, --dry-run writing nothing, and the STATE.md
reference append. Full clean rebuild + npm test: 15/15 passing.

### DIFF
--- a/tools/loop-sync/README.md
+++ b/tools/loop-sync/README.md
@@ -31,8 +31,8 @@ loop-sync [target-dir] [options]
 
 | Option | Description |
 |--------|-------------|
-| `-a, --auto-fix` | Attempt to auto-fix issues (experimental) |
-| `-d, --dry-run` | Show what would be done without making changes |
+| `-a, --auto-fix` | Scaffold missing STATE.md, gate.yaml, loop-budget.md, and loop-run-log.md with minimal defaults, and add a missing STATE.md reference to LOOP.md. LOOP.md and AGENTS.md are never fabricated — run `loop-init` for those. |
+| `-d, --dry-run` | With `--auto-fix`, report what would be scaffolded/changed without writing anything |
 | `-v, --verbose` | Show detailed information |
 | `--json` | Output JSON format |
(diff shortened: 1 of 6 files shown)


cobusgreyling (approved):
Implements the documented --auto-fix/--dry-run flags that were previously no-ops. Scope is right: only pattern-agnostic files are scaffolded, LOOP.md/AGENTS.md stay report-only, dry-run writes nothing. Tests cover the four cases I care about.

PR #524, opened 2026-08-15
fix(loop-context): don't crash when an --on-exceed hook ignores stdin

KhaiTrang1995:
## Problem
runOnExceedHook() piped the escalation decision to the hook script's
stdin with no 'error' listener on the stream itself. If the script
exits without ever reading stdin -- a completely ordinary shape for a
notify script (e.g. one that just does `echo escalated`) -- writing to
that already-closed pipe raises an 'error' event on the stream (EPIPE
on POSIX, EOF on Windows). Node treats an unhandled stream 'error' as
an uncaught exception by default, crashing the whole loop-context
process, even though this hook is documented and coded as
fire-and-forget: the crash happens after the circuit breaker has
already made its decision, so it turns a successful --check run into
a hard failure for a reason entirely unrelated to the ledger.

## Fix
Attach a no-op 'error' listener on child.stdin so a write-after-close
failure is absorbed instead of escalating past the 'close' handler,
which already resolves the promise regardless of how the hook exited.

## Test plan
Added a regression test with a hook script that never reads stdin,
using a large ledger error string so the piped decision JSON reliably
exceeds the OS pipe buffer (removing timing flakiness from the repro)
-- confirmed it fails with the exact "write EOF" crash against the
pre-fix code and passes with the fix. Full clean rebuild + npm test:
52/52 passing.

### DIFF
--- a/tools/loop-context/dist/cli.js
+++ b/tools/loop-context/dist/cli.js
@@ -276,6 +276,14 @@ function runOnExceedHook(script, decision) {
             resolve();
         });
         child.on('close', () => resolve());
+        // The hook script may exit without ever reading stdin (e.g. a notify
+        // script that ignores its input). Writing to that already-closed pipe
+        // then raises an 'error' event (EPIPE on POSIX, EOF on Windows) on the
+        // stream itself; left unhandled, Node treats that as an uncaught
+        // exception and crashes the whole process. This hook is documented as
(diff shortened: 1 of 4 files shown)


cobusgreyling (approved):
Fire-and-forget --on-exceed hook must not crash the process on EPIPE when the script never reads stdin. No-op stdin error listener is the right fix; the large-payload test makes the race reliable.

PR #525, opened 2026-08-15
fix(readiness-core): don't let an unreadable skills path abort the whole scan

KhaiTrang1995:
## Problem
scanSkillDirectories() only guarded the readdir() call with a prior
fileExists() check, which just proves the path existed at stat() time
-- it doesn't prove readdir() will succeed on it. A skills path that
exists but isn't actually a directory (renamed, a broken checkout, a
stray file artifact) throws ENOTDIR; an unreadable or permission-
denied directory throws EPERM; a path removed between the two calls
throws ENOENT. All three used to propagate out of this function
uncaught, which aborts the entire audit run in loop-audit and
goal-audit (both call this via findSkills) instead of just skipping
that one broken path and scanning the rest.

## Fix
tools/mcp-server/src/resolver.ts's listSkills() already handles the
identical shape of problem correctly, wrapping its readdir() in a
try/catch with the same fileExists()-then-readdir() structure. Applied
the same fix here.

## Test plan
Added a regression test where .claude/skills exists as a file instead
of a directory, asserting the scan completes and still finds a valid
skill under a sibling path instead of throwing. Full clean rebuild +
npm test: 6/6 passing.

### DIFF
--- a/tools/readiness-core/src/index.ts
+++ b/tools/readiness-core/src/index.ts
@@ -39,10 +39,19 @@ export async function scanSkillDirectories(root: string): Promise<string[]> {
   const foundSet = new Set<string>();
   for (const dir of dirs) {
     if (!(await fileExists(dir))) continue;
-    const entries = await readdir(dir, { withFileTypes: true });
-    for (const e of entries) {
-      if (e.isDirectory()) foundSet.add(e.name);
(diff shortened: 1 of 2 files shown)


cobusgreyling (approved):
scanSkillDirectories should skip unreadable/non-dir skills paths instead of aborting the whole audit. Matches mcp-server listSkills() and the ENOTDIR fixture test is good.

PR #529, opened 2026-08-17
 fix(loop): guard the top-level --interactive shortcut with a TTY check

KhaiTrang1995:
## Problem
cmdWizard() (loop wizard --interactive) already gates the interactive
path on process.stdin.isTTY, falling back to printNonInteractiveHelp()
otherwise. The top-level `loop --interactive` shortcut in main() never
had the same guard: it called runInteractiveWizard() unconditionally.

readline's question() resolves immediately with '' when stdin isn't a
TTY (a CI step, piped input, `< /dev/null`) instead of waiting for a
human. Without the guard, `loop --interactive` in a non-interactive
context silently ran the wizard with every prompt answered blank, then
executed the resulting default plan -- `loop init . --pattern
daily-triage --tool grok` against the current directory -- scaffolding
files nobody asked for instead of printing help.

## Fix
Mirrors cmdWizard()'s existing gate at the other --interactive entry
point.

## Test plan
Added a regression test that runs `loop --interactive` with stdin
closed (execFileSync's input: '', matching `< /dev/null`) in an empty
temp directory, asserting it prints the non-interactive help and the
directory stays empty. Confirmed it fails with the wizard prompts
printed and no help text when the guard is stripped, and passes with
the fix. Full clean rebuild + npm test: 11/11 passing.

### DIFF
--- a/tools/loop/dist/cli.js
+++ b/tools/loop/dist/cli.js
@@ -139,6 +139,17 @@ async function main() {
         return;
     }
     if (cmd === '--interactive' || cmd === '-i') {
+        // Mirrors cmdWizard()'s TTY gate below: runInteractiveWizard() drives
+        // readline.question(), which resolves immediately with '' when stdin
+        // isn't a TTY (a CI step, piped input, `< /dev/null`) instead of
+        // waiting for a human. Without this guard, loop --interactive in a
+        // non-interactive context silently ran the wizard with every prompt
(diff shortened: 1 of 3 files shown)


cobusgreyling (approved):
Mirrors the existing cmdWizard() TTY gate. Non-TTY loop --interactive was silently running the default init plan; help + no scaffold is the right behavior. Regression test with closed stdin is solid.

PR #533, opened 2026-08-17
fix(mcp-server): stop mislabeling every estimateCost failure as an invalid cadence

KhaiTrang1995:
## Problem
The loop_estimate_cost tool's catch-all around estimateCost() always
reported "Invalid cadence: <value>" regardless of what actually threw.
estimateCost() can fail for reasons that have nothing to do with
cadence -- e.g. a registry.yaml pattern entry with a missing or
malformed `cost` block throws a TypeError reading its fields. Anyone
debugging a broken registry entry got pointed at the wrong cause
entirely.

## Fix
Surface the real thrown message instead of guessing.

## Test plan
Two new tests -- one confirms a genuinely bad cadence override still
reports loop-cost's own "Invalid cadence" message (not a hardcoded
guess), the other adds a pattern missing its `cost` block and asserts
the resulting error does not claim the cadence is invalid. Confirmed
the second one fails with the old catch-all ("Invalid cadence: 1d" for
a perfectly valid cadence) and passes with the fix. Full clean rebuild
+ npm test: 28/28 passing.

### DIFF
--- a/tools/mcp-server/dist/index.js
+++ b/tools/mcp-server/dist/index.js
@@ -284,8 +284,15 @@ server.tool('loop_estimate_cost', 'Estimate daily token cost for a pattern at a 
     try {
         result = estimateCost({ pattern, level, cadence });
     }
-    catch {
-        return { content: [{ type: 'text', text: `Invalid cadence: ${cadence ?? pattern.cadence}` }] };
+    catch (err) {
+        // estimateCost() can fail for reasons that have nothing to do with
+        // cadence -- e.g. a registry.yaml entry with a missing/malformed
+        // `cost` block throws a TypeError reading its fields. Hardcoding
+        // "Invalid cadence" here regardless of the actual failure misled
+        // whoever was debugging a broken registry entry. Surface the real
+        // message instead.
+        const message = err instanceof Error ? err.message : String(err);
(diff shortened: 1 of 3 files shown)


cobusgreyling:
Needs review: `validate` and `audit` are green and the PR is mergeable, but this changes MCP `loop_estimate_cost` error text (and the committed `dist/`). Not merging from housekeeping — please take a quick maintainer look.

KhaiTrang1995:
Reviewed as MCP maintainer. Confirmed:
- dist/index.js is byte-identical to a clean `tsc` build from src/index.ts (no manual drift).
- 28/28 tests pass on a full clean rebuild.
- No other file in the repo depends on the old "Invalid cadence:" text.
- release-loop-mcp-server.yml only publishes on a version tag, so merging
  this doesn't auto-publish to npm; package.json stays at 1.2.0.

The error-text change is intentional and correct (see PR description) —
approving for merge.

cobusgreyling (approved):
Good catch. Swallowing every \`estimateCost\` throw as “Invalid cadence” hid real registry errors (missing \`cost\` block → TypeError). Surfacing \`err.message\` is the right MCP behavior.

Tests cover both a genuine bad cadence (still matches /Invalid cadence/i via loop-cost’s own message) and a broken registry entry (must not blame cadence). \`dist/index.js\` matches source. Approving.

PR #536, opened 2026-08-18
fix(safety): anchor gate.yaml denylist patterns so nested sensitive paths are actually caught

KhaiTrang1995:
## Problem
Every shipped denylist -- the repo's own gate.yaml, docs/safety.md's
prose (the source loop-gate is documented as mirroring), the
templates/gate.yaml.template a new project copies, and the gate.yaml
generated by both loop-audit --fix and loop-sync --auto-fix -- listed
bare, unanchored patterns for the plain-filename and directory-style
entries: ".env", ".env.*", ".terraform/**", "k8s/production/**",
"auth/**", "payments/**", "billing/**".

tools/loop-gate matches with plain minimatch(path, glob, { dot: true
}), no matchBase. A bare pattern like ".env" only matches a path
*exactly equal to* ".env" -- it does not match ".env" nested in any
subdirectory. Verified directly against minimatch:

  minimatch('services/api/.env', '.env', {dot:true})              -> false
  minimatch('services/auth/x.ts', 'auth/**', {dot:true})          -> false
  minimatch('apps/k8s/production/deploy.yaml', 'k8s/production/**', {dot:true}) -> false

Concrete failure: `loop-gate check --action commit --paths
services/api/.env` -- a completely ordinary monorepo layout -- comes
back "Within policy — cleared to proceed." The credential file
silently passes the exact mechanism docs/safety.md says "enforces this
denylist... mechanically... instead of relying on the loop to have
read this file." The same gap applies to any nested auth/, payments/,
billing/, k8s/production/, or .terraform/ directory.

## Fix
Prefixed every bare/directory-style entry with **/ (already a no-op
superset for root-level matches -- verified .env and **/.env match
identically at repo root, **/.env additionally matches every nested
depth). The entries that already carried **/ (secrets, credentials,
*_key*, *_secret*, migrations) were correct as shipped.

Also fixes the same bare patterns baked into tools/loop-sync's own
gate.yaml auto-fix scaffold, since it copied the same denylist.

## Test plan
Added a regression test in tools/loop-gate that loads this repo's
actual gate.yaml and asserts nested paths under every denylist
category are caught -- confirmed it fails with "Within policy" against
the pre-fix content and passes with the fix. Strengthened
tools/loop-audit's and tools/loop-sync's existing scaffold tests to
assert every generated denylist entry is anchored, closing the gap
where the prior test only checked the written gate.yaml was
schema-valid, never that loop-gate's matcher actually catches a nested
path with it. Full clean rebuild + test across loop-gate (32/32),
loop-audit (26/26), and loop-sync (15/15).

### DIFF
--- a/docs/safety.md
+++ b/docs/safety.md
@@ -7,18 +7,18 @@ Loops amplify judgment — good and bad. These guardrails are minimum bar for pr
 The loop must **never** auto-edit these without human approval:
 
 ```
-.env
-.env.*
+**/.env
+**/.env.*
 **/secrets/**
 **/credentials/**
 **/*_key*
 **/*_secret*
-.terraform/**
-k8s/production/**
+**/.terraform/**
+**/k8s/production/**
 **/migrations/**          # unless explicit migration loop
-auth/**
-payments/**
-billing/**
+**/auth/**
+**/payments/**
+**/billing/**
 ```
 
 Encode in `minimal-fix` and implementer skills:
--- a/gate.yaml
+++ b/gate.yaml
@@ -4,18 +4,18 @@
 version: 1
 
 denylist:
-  - ".env"
-  - ".env.*"
+  - "**/.env"
+  - "**/.env.*"
   - "**/secrets/**"
   - "**/credentials/**"
   - "**/*_key*"
   - "**/*_secret*"
-  - ".terraform/**"
-  - "k8s/production/**"
+  - "**/.terraform/**"
+  - "**/k8s/production/**"
   - "**/migrations/**"
-  - "auth/**"
-  - "payments/**"
-  - "billing/**"
+  - "**/auth/**"
+  - "**/payments/**"
+  - "**/billing/**"
 
 maxFiles: 10
 
--- a/templates/gate.yaml.template
+++ b/templates/gate.yaml.template
@@ -5,18 +5,18 @@
 version: 1
 
 denylist:
-  - ".env"
-  - ".env.*"
+  - "**/.env"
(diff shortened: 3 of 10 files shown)


cobusgreyling:
Needs review: this retargets `gate.yaml` denylist matching (and templates/sync/audit). Safety-path change — not merging from housekeeping even though `validate`/`audit` are green.

KhaiTrang1995:
ThanksThis only broadens denylist matching (adds **/
prefixes) -- denylist is checked before allowlist/maxFiles in gate.ts
and always wins, so this change can only cause more escalations, never
let something previously blocked slip through.

Confirmed:
- dist/ is content-identical to a clean rebuild across loop-gate,
  loop-audit, and loop-sync (only CRLF/LF noise from a Windows
  checkout, no real drift).
- 32/32, 26/26, 15/15 tests pass on full clean rebuilds.
- No path in this repo (outside node_modules) becomes newly
  denylisted by the change.

cobusgreyling (approved):
Verified against minimatch (\`dot: true\`, same as \`loop-gate\`): \`**/.env\`, \`**/auth/**\`, etc. match both repo-root paths (\`.env\`, \`auth/session.ts\`) *and* nested ones (\`services/api/.env\`). So this is a real coverage fix, not a root-path regression.

Templates, autofixer, and loop-sync stay in lockstep with \`gate.yaml\` / \`docs/safety.md\`. The dogfood test against this repo’s own \`gate.yaml\` is the right assertion. Approving.

PR #537, opened 2026-08-18
 fix(loop-metrics): stop successRatePct from going negative on multi-escalation runs

KhaiTrang1995:
## Problem
successRatePct was computed as ((totalRuns - totalEscalations) /
totalRuns) * 100. totalEscalations is a sum of each run's escalation
*count* (a run can log more than one -- e.g. several items escalated
in a single triage pass; this repo's own loop-run-log.md has real
entries with escalations: 4 and escalations: 5), not a count of
runs-that-escalated. Subtracting an event count from a run count
conflates the two: any window where totalEscalations exceeds
totalRuns (trivially reachable with real data) produces a negative
percentage, which the CLI dashboard prints and color-codes as-is with
no clamping.

## Fix
Success rate should mean "what fraction of runs didn't escalate at
all." Count runs with escalations === 0 instead of subtracting the
raw event sum, so the result stays correctly bounded to [0, 100]
regardless of how many escalations any single run logged.
totalEscalations itself is unchanged since roiScore correctly treats
each individual escalation event as a cost.

## Test plan
Added a regression test with one run logging escalations: 5 and one
clean run, asserting successRatePct is 50 (not -150). Confirmed it
fails with the exact -150 result against the old formula and passes
with the fix. Full clean rebuild + npm test: 3/3 passing.

### DIFF
--- a/tools/loop-metrics/dist/metrics.js
+++ b/tools/loop-metrics/dist/metrics.js
@@ -48,14 +48,25 @@ export function aggregateMetrics(entries) {
     let totalDurationS = 0;
     let totalActionsTaken = 0;
     let totalEscalations = 0;
+    let runsWithoutEscalation = 0;
     for (const entry of entries) {
         totalTokens += entry.tokens_estimate || 0;
         totalDurationS += entry.duration_s || 0;
         totalActionsTaken += entry.actions_taken || 0;
         totalEscalations += entry.escalations || 0;
+        if (!entry.escalations)
+            runsWithoutEscalation++;
     }
     const totalRuns = entries.length;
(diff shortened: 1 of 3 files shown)


cobusgreyling:
Needs review: `validate`/`audit` are green, but this changes how `successRatePct` is defined (run-level “no escalations” vs subtracting total escalation *events*). That is a metrics semantics change, not a trivial fix — not merging from housekeeping.

cobusgreyling (approved):
The old formula subtracted an *event* count (\`totalEscalations\`) from a *run* count, so a single run with 5 escalations could push \`successRatePct\` negative. Treating success as “runs with zero escalations” is the right definition and stays in [0, 100].

Tests cover the existing 50% case and the multi-escalation-in-one-run case. \`dist/metrics.js\` matches \`src\`. Fine to land.

PR #599, opened 2026-09-10
feat(loop-sync): detect gate.yaml ↔ docs/safety.md policy drift

THRISHAL12345:
## Problem

`gate.yaml` describes itself as the machine-readable twin of `docs/safety.md`'s Path Denylist:

```yaml
# Machine-readable twin of docs/safety.md's Path Denylist and Auto-Merge
# Policy sections. Used by tools/loop-gate. Keep in sync with docs/safety.md
# when either changes.
```

That "keep in sync" is done by hand, and nothing verifies it. In a project whose thesis is *mechanical enforcement over trusting an agent to have read a markdown file*, this was the one invariant left on the honor system.

The dangerous direction is silent. Add a path to `docs/safety.md` but forget `gate.yaml`, and:

- humans read `docs/safety.md` and believe the path is protected
- `loop-gate` never sees it and keeps exiting `0` on changes that touch it
- every downstream project that copied the pattern inherits the same hole

## What this does

Extends `loop-sync` — which already owns config drift and already reads `gate.yaml` as a required file — to compare the two policies on every run.

| Divergence | Severity | Rationale |
|---|---|---|
| Documented in `safety.md`, **not enforced** by `gate.yaml` | `error` | Silent hole; the doc lies to readers |
| Enforced by `gate.yaml`, **undocumented** in `safety.md` | `warning` | Safe, but invisible to humans |
| `gate.yaml` limits (`maxFiles`) the prose never states | `info` | Discoverability |

Design decisions worth reviewing:

- **Read-only, even under `--auto-fix`.** Which side of a safety divergence is correct is a human decision. `AGENTS.md` already keeps safety surfaces under human review, so this reports and never rewrites. There is a test asserting `--auto-fix` leaves `gate.yaml` byte-identical.
- **Skipped when either file is absent.** A project with no `docs/safety.md` has no prose policy to drift from, so this stays quiet rather than nagging.
- **Exact string comparison** on trimmed globs. `**/auth/**` and `**/auth/*` are semantically close but should not silently count as agreeing — for a safety policy the doc and the enforced file should read identically.
- **Extends `runSync` rather than adding a package**, keeping the front door and `loop doctor` unchanged.

## The first real finding

Run against this repo, the check immediately surfaced one: `gate.yaml` sets `maxFiles: 10` — an escalation threshold documented nowhere. Denylist entries were already in sync; the limit was not. Documented in `docs/safety.md` in this PR, taking the repo from 89 → 90 healthy.

## Verification

- **36/36 loop-sync tests pass** (16 new), clean `npm ci` from the updated lockfile.
- **Live drift proof** — removing `**/payments/**` from the real `gate.yaml` drops the score 89 → 69 and raises the error; restored clean.

```
❌ Errors:
   - gate.yaml ↔ docs/safety.md: 1 denylist path documented in docs/safety.md
     but not enforced by gate.yaml: **/payments/**
```

- **`loop-gate` (32) and `loop` front door (11) still pass** — `doctor` consumes sync's JSON, so that consumer was checked.
- **No CI change needed.** `scripts/ci-validate-gates.sh` already runs loop-sync's tests, and one new test asserts *this repo's* `gate.yaml` and `docs/safety.md` agree — so drift now fails the build.

## One call to flag

This adds `yaml@^2.8.0` (the version `loop-gate` already pins), which ends `loop-sync`'s zero-dependency status. I judged correct YAML parsing worth more than dependency purity for a file that gates production paths, but I'm happy to hand-roll a narrow parser instead if you'd rather keep the package dep-free.


cobusgreyling (approved):
Reviewed the policy-drift work. This is the right owner (`loop-sync` already reads `gate.yaml`) and the severity split is correct: documented-but-unenforced is an error (the doc lies; `loop-gate` would allow the path), undocumented enforcement is a warning, `maxFiles` is info.

Also checked:
- Comparison is exact glob string equality, which is what we want for a safety twin rather than “semantically overlapping.”
- `--auto-fix` does not rewrite either file (test covers this).
- Parser handles inline `#` comments, CRLF, missing/unterminated fences, and malformed YAML without throwing.
- `docs/safety.md` now states `maxFiles: 10` so this repo’s own dogfood check is green.

Approving. CI (`validate`, `audit`) is green and the PR is mergeable.

PR #610, opened 2026-09-15
fix(loop-metrics): preserve non-date run IDs in timeframe filters

Iams4kura:
## Summary

`loop-metrics --timeframe 7d` silently drops a run whose ID is `run-1`, reporting **0 runs / 0 tokens** for a log that contains **1 run / 52k tokens**. JavaScript accepts that opaque ID as a date in 2001, so the existing `Invalid Date` check never preserves it. Short numeric IDs such as `123` and `2026` have the same problem.

Only interpret ISO-shaped run IDs as dates, matching the existing run-log pruner. IDs without a usable timestamp remain visible; actual old ISO dates are still filtered out.

## Related issue

Follow-up to merged #499 and #519. #499 preserves IDs that produce `Invalid Date`; this covers opaque IDs that JavaScript *successfully* parses as dates. #519 already applies the ISO-shape guard to `scripts/append-run-log.mjs`, but not to the metrics reader. Both changes are present in the tested base `ed0c1a29b106e85ed568a1241b916161e05f4a53`.

## Changes

- [ ] Doc / example / story improvement
- [x] Tool / CLI change under `tools/`
- [ ] Test only
- [ ] Other

The patch adds an ISO-shape guard to `filterEntries`, synchronizes the tracked package runtime, and extends the existing timeframe regression with opaque IDs, recent/old/invalid ISO dates and pattern filtering. The regression uses a fixed clock so its July fixture does not expire with wall time.

- [x] Links work from README or the relevant index — no navigation links changed
- [x] No secrets, tokens, or internal company URLs
- [x] `STATE.md*` examples use `.example` suffix — no state examples changed
- [ ] Safety-related content references `docs/safety.md` — not applicable; report filtering only
- [x] If you touched a package under `tools/<pkg>`: `cd tools/<pkg> && npm ci && npm test`
- [ ] If you touched the registry: root `npm ci && npm run validate:registry` — registry unchanged; CI validation still run

## Testing / Dogfood

- [x] `npx @cobusgreyling/loop audit .` (or package-local tests) on affected paths
- [ ] Manual review of generated state / skill output if scaffolding changed — no scaffolding changed

- On the unchanged base, the final regression fails because `run-1`, `123`, and `2026` are missing. The same test passes after the fix.
- `cd tools/loop-metrics && npm test`: **3 passed**, including TypeScript build.
- Independent verification: **3 passed** on Node 18.20.8 and Node 20.20.2 (`TZ=America/New_York`); cutoff boundary remains inclusive, while one second older is excluded.
- `bash scripts/ci-validate-gates.sh`: passed.
- `bash scripts/ci-audit-gates.sh`: passed; reference score 100. The scripts accept their documented doctor warning and unconfigured token-budget drill skip.
- `git diff --cached --check`: passed.
- Real CLI, same one-entry `run-1` log with `--timeframe 7d`: **0 runs / 0 tokens → 1 run / 52k tokens**.

---

Docs, stories, adopters, and small tests: **maintainers aim to review within 48 hours** (same-day when possible).


cobusgreyling (approved):
Looks correct. `new Date("run-1")` is a valid (ancient) date, so a days-window filter was dropping non-ISO run IDs. Gating on `/^\d{4}-\d{2}-\d{2}/` matches the run-log pruner and the tests cover `run-1`, numeric GitHub ids, slugs, and invalid `2026-99-99`. CI green. Approving.

